PR #176, opened 2020-10-28
Issue 166: Current offset for byte stream writer

### LINKED ISSUE #166: Implement Seek for ByteStreamWriter
**Problem description**
Although a Pravega stream is append-only, implementing `Seek` for `ByteStreamWriter` would allow one to use the standard file-like API `std::io::Seek(SeekFrom::Current(0))` to obtain the offset of the next append. Although this is not a very elegant API, it is a standard way to obtain the current position in a file. Note that there is a new method `Seek::stream_position()` but this is only available in nightly builds of Rust (https://doc.rust-lang.org/std/io/trait.Seek.html#method.stream_position).

**Problem location**
byte_stream.rs

I implemented this here:
https://github.com/claudiofahey/pravega-client-rust/blob/097a286c9be4619457035b5fa975fee81ef336ba/src/byte_stream.rs#L168

Tristan1900:
**Change log description**  
Byte stream writer maintains a write_offset that is used for conditional append. If conditional append failed when calling `flush`, user can deicide to panic or call `seek_to_tail` to fetch the latest write_offset of that segment and continue to do the append.

**Purpose of the change**  
Fix #166


src/byte_stream.rs
```diff
@@ -112,14 +114,39 @@ impl ByteStreamWriter {
             sender,
             metadata_client,
             runtime_handle: handle,
-            event_handle: None,
+            event_handles: vec![],
+            current_offset: 0,
```
tkaitchuck, line 118: This is not strictly correct. It is possible that the client is opening on an existing segment.

tkaitchuck (changes requested):

src/byte_stream.rs
tkaitchuck: I think we can use an option and do the lazily.

Tristan1900: Thanks, I see. BTW,  I realize that we can just send a wirecommand to the server and get the offset without maintaining anything in the bytestream writer.

claudiofahey: I think there is confusion about the original request. The overall requirement for the GStreamer use case is that the application needs to know the offset that a write to will go to. This offset will be used in the index. There are a few ways of achieving this:
Option 1 (preferred, in my opinion): The client tracks the offset. stream_position() returns this client-tracked offset. write() calls are implemented as conditional appends to the expected offset. In the unexpected case that another instance of ByteStreamWriter appends data to the same stream, this conditional append will return an error and the client can choose to panic or seek to the tail.
Option 2: The client tracks the offset. stream_position() returns this client-tracked offset. write() calls are normal appends.   In the unexpected case that another instance of ByteStreamWriter appends data to the same stream, the client-tracked offset will be incorrect forever.
Option 3: The client does not track the offset. stream_position() queries the server for the current tail and is expensive. write() calls are normal appends. In the unexpected case that another instance of ByteStreamWriter appends data to the same stream, then stream_position() may or may not return the offset of the next append (this is a race condition).

Tristan1900: Sounds like in this case we need to use `ConditionalAppend` wirecommand and have something similar to the `ConditionalOutputStream` .

tkaitchuck: @Tristan1900 Yes. I think we can probably have a either a flag or a "wireCommandFactory" passed into `SegmentReactor` which the EventStream sets to `AppendBlockEnd` and the ByteStream sets to `ConditionalAppend`

src/reactor/segment_writer.rs
```diff
@@ -239,7 +263,7 @@ impl SegmentWriter {
                 SegmentWriter::MAX_WRITE_SIZE
             );
             if append.event.data.len() + to_send.len() <= SegmentWriter::MAX_WRITE_SIZE as usize
-                || event_count == SegmentWriter::MAX_EVENTS as usize
+                && event_count < SegmentWriter::MAX_EVENTS as usize
```
tkaitchuck, line 266: While none of the current paths do this, we should handle that conditional and non-conditional appends are mixed, as this is logically possible given the API of this calss.

The logic here should be that conditional appends can be combined if their conditional lengths presume the prior append is a success. However conditional and non-conditional ones would need to be separate variables. 
So the condition should be added to this `if` so that is breaks out of the loop if it switches from conditional to non-conditional.

tkaitchuck: On line 270 below, it could be:
`to_send.extend(&append.event.data);`
which would avoid making a copy.

Tristan1900: Thanks! I have updated the code to separate the conditional append with the regular append.

Tristan1900:
This PR is depending on [this](https://github.com/pravega/pravega/pull/5301) to work.

tkaitchuck (approved):

Tristan1900:
@claudiofahey This PR now is depending on [this PR](https://github.com/pravega/pravega/pull/5319). Instead of adding a new wirecommand, we decide to modify the existing one. We probably should not merge this PR until that modification is merged into Pravega.

PR #188, opened 2020-11-06
Issue 186: Ensure RUST logs are bridged to Python.

### LINKED ISSUE #186: Ensure RUST client logs are visible while executing Python Bindings.
**Problem description**
The native client logs are not visible when the Python bindings are executed.

**Problem location**
Logging

Use pyo3-log crate https://crates.io/crates/pyo3-log to enable this.

shrids:
This approach is still under discussion and evaluation.

**Change log description**  
- Use pyo3-log crate to bridge RUST logs to Python.

**Purpose of the change**  
Fixes #186 

**What the code does**  
- Enable pyo3-log crate. (version 0.1.x is used since we are using pyo3 0.11.x. )
- Ensure "tracing" crate is used instead of log crate.
- Uses `py.allow_threads` API from pyo3 to ensure RUST logs are returned.
This API temporarily releases the GIL, thus allowing other Python threads to run.
Documentation: https://docs.rs/pyo3/0.11.1/pyo3/struct.Python.html#method.allow_threads

(At the moment this change has been enabled only for the `create_scope ` python API . It will be enabled to other APIs based on discussions/evaluation...)

Usage pattern:
```python
import logging
logging.getLogger().setLevel(logging.DEBUG) # Initialize logging level
FORMAT = '%(levelname)s %(name)s %(asctime)-15s %(filename)s:%(lineno)d %(message)s'
# Configure the log format and filename.
logging.basicConfig(filename="pravega_debug.log", level=logging.DEBUG, filemode="w+", format=FORMAT)
logging.info("") # this inits the logging.
# pravega client needs to be imported after the logging is initialized.
import pravega_client

logging.info("Logging from Python")
stream_manager=pravega_client.StreamManager("127.0.0.1:9090")
stream_manager.create_scope("sc1")
```

**How to verify it**  
All the existing tests pass.
`pravega_reader_test.py `- This file shows an example usage. The log files are persisted in file `pravega_debug.log`
Both the python and RUST logs are logged here.


shrids:
Closing this PR since there is not way by which we can enable all the RUST client logs from the python. The method of using API
```
py.allow_threads(...)
```
to enable RUST logs in python is restrictive i.e. we will not be able to reuse the writer instance to write multiple events.
Closing this PR since https://github.com/pravega/pravega-client-rust/pull/206 is more cleaner solution.

PR #189, opened 2020-11-09
Issue 180: Prefetch read

### LINKED ISSUE #180: Add prefetch to AsyncSegmentReader (via wrapper)
**Problem description**
Instead of #153 which does not affect events we should take advantage of the fact that in `AsyncSegmentReader` reads are highly predictable. 

**Problem location**
In AsyncSegmentReader::read each time the call is made a new request it issued to the controller to locate the endpoint, and then upon completion it returns just the data that was requested. This results in two round trip latencies on call paths that are perceptible to the end user.

We should create a PrefetchingAsyncSegmentReader which wraps AsyncSegmentReader. This will issue larger read calls than the caller requested and hold onto the data locally. If the next call is for the following offset it can simply be returned without having to fetch more data. (Or wait on the future of the already issued request) We don't have to worry about boundaries because it is valid to return less data than the caller asked for.

Additionally AsyncSegmentReader should be modified to be optimistic and assume that the endpoint has not changed on the first attempt and only fetch the new endpoint when an error is encountered.

Tristan1900:
**Change log description**  
Add prefetch for async segment reader.

**Purpose of the change**  
Fix #180 

**What the code does**  
Add a AsyncSegmentReaderWrapper that maintains an internal buffer and a read offset. Every time read request comes in it will fetches data from the buffer and issues a background read request to server if needed  to fill the buffer. Basically this wrapper will enable caller to read from memory instead of waiting for server to response.


src/segment_reader.rs
```diff
@@ -342,6 +345,142 @@ impl AsyncSegmentReaderImpl {
...
+    reader: Arc<Box<dyn AsyncSegmentReader>>,
+    offset: i64,
+    end_of_segment: bool,
+    is_truncated: bool,
+    outstanding: Outstanding,
+    handle: Handle,
+}
+
+impl AsyncSegmentReaderWrapper {
+    const DEFAULT_BUFFER_SIZE: usize = 1024 * 1024;
```
claudiofahey, line 363: Allow these buffer parameters to be specified by the user at runtime.

Tristan1900: got it

integration_test/src/byte_stream_tests.rs
```diff
@@ -145,9 +153,11 @@ fn test_seal(writer: &mut ByteStreamWriter, reader: &mut ByteStreamReader, rt: &
     assert_eq!(size, 4);
     assert_eq!(buf, vec![2; 4]);
 
+    error!("read here");
```
tkaitchuck, line 156: ?

src/segment_reader.rs
```diff
@@ -342,6 +346,157 @@ impl AsyncSegmentReaderImpl {
     }
 }
 
+/// A wrapper around the AsyncSegmentReader.
+///
+/// It maintains a buffer and prefetches data to fill that buffer in the background so that
+/// every read call can read from the local buffer instead of waiting for server response.
+pub(crate) struct AsyncSegmentReaderWrapper {
```
tkaitchuck, line 353: The name `AsyncSegmentReaderWrapper` implies that it's a generic wrapper of `AsyncSegmentReader` while it is a wrapper, it provides some additional functionality, it should be named accordingly. Like maybe `BufferedAsyncSegmentReader` or `PrefetchingAsyncSegmentReader`.

src/segment_reader.rs
```diff
@@ -342,6 +346,157 @@ impl AsyncSegmentReaderImpl {
...
+
+    pub(crate) fn extract_reader(self) -> Arc<Box<dyn AsyncSegmentReader>> {
+        self.reader
+    }
+
+    fn issue_request_if_needed(&mut self) {
+        let updated_read_length = cmp::max(self.read_size, self.buffer.capacity() - self.buffer.len());
+        if !self.end_of_segment && !self.is_truncated && self.outstanding.is_empty() {
+            let (sender, receiver) = oneshot::channel();
+            self.handle.enter(|| {
```
tkaitchuck, line 413: I think `handle.enter(|| tokio::spawn( ...` can also be written as just `handle.spawn(...`

src/segment_reader.rs
```diff
@@ -342,6 +346,157 @@ impl AsyncSegmentReaderImpl {
...
+            .map_err(|e| error!("failed to send reply back: {:?}", e))
+            .expect("send reply back");
+    }
+
+    async fn fill_buffer(&mut self) -> Result<(), ReaderError> {
+        let data = self.outstanding.get_data().await?;
+
+        let to_fill = cmp::min(self.buffer.capacity() - self.buffer.len(), data.len());
+        self.buffer.extend(&data[..to_fill]);
+        data.drain(0..to_fill);
```
tkaitchuck, line 448:
```suggestion
        self.buffer.extend(data.drain(0..to_fill).as_slice());
```

src/segment_reader.rs
```diff
@@ -342,6 +346,157 @@ impl AsyncSegmentReaderImpl {
...
+        };
+        wrapper.issue_request_if_needed();
+        wrapper
+    }
+    pub(crate) async fn read(&mut self, buf: &mut [u8]) -> StdResult<usize, ReaderError> {
+        self.issue_request_if_needed();
+        while self.buffer.is_empty() {
+            if self.outstanding.is_empty() {
+                return Ok(0);
+            }
```
tkaitchuck, line 393: I feel like this is mixing paradigms. 
It's an async method. I think it's OK for it to return a result only after the operation is done.
So it probably makes more sense to only return once the requested data is available.

src/segment_reader.rs
```diff
@@ -342,6 +346,157 @@ impl AsyncSegmentReaderImpl {
     }
 }
 
+/// A wrapper around the AsyncSegmentReader.
+///
+/// It maintains a buffer and prefetches data to fill that buffer in the background so that
+/// every read call can read from the local buffer instead of waiting for server response.
+pub(crate) struct AsyncSegmentReaderWrapper {
+    buffer: Vec<u8>,
```
tkaitchuck, line 354: Random thought: What if instead of having a vec of bytes that we copy replies out of, we just held a vec of replies?

src/segment_reader.rs
Tristan1900: fixed, thanks!

src/segment_reader.rs
Tristan1900: got it!

src/segment_reader.rs
Tristan1900: That's a very good idea. I have updated the implementation.

src/segment_reader.rs
```diff
@@ -342,6 +348,157 @@ impl AsyncSegmentReaderImpl {
...
+                // no more bytes could be read from the current offset
+                return Ok(0);
+            }
+            self.fill_buffer().await?;
+            self.issue_request_if_needed();
+        }
+
+        let mut need_to_read = buf.len();
+        let mut copy_offset = 0;
+        while let Some(mut cmd) = self.buffer.pop_back() {
```
tkaitchuck, line 406: According the the docs for VecDeque:
> The "default" usage of this type as a queue is to use push_back to add to the queue, and pop_front to remove from the queue.

I don't know if this is actually any faster. But as this direction is the "default" we should prefer it all else being equal.

tkaitchuck (approved):

PR #193, opened 2020-11-24
Issue 192: Keycloak support

### LINKED ISSUE #192: Support keycloak
**Problem description**
Auth model need to support keycloak integration

**Problem location**
Auth

Tristan1900:
**Change log description**  
Support keycloak

**Purpose of the change**  
Fix #192 

**What the code does**  
If auth is enabled and keycloak is chosen, client will first try to read a `keycloak.json` file. After that Pravega Rust client will issue two POST requests to the keycloak server, the first POST request is to obtain an access token, so the Pravega Rust client can be authenticated and becomes a keycloak client. The second POST request is to use the access token to ask for permissions on behalf a user, which is the Rust client itself. The token returned from the second POST is called RPT(relying party token), with this token anyone can access to the given resources, which in our case, is just the Pravega controller. The RPT token is used by controller client in Rust client for authentication and authorization.

**How to verify it**  
Starts a SDP cluster and runs an application that uses the Pravega Rust client.


config/src/credentials.rs
```diff
@@ -37,26 +41,152 @@ impl Credentials {
...
     }
+
+    #[test]
+    fn test_json_deserialize() {
+        let json_string = r#"{"realm":"nautilus","auth-server-url":"http://keycloak.jarviscb.nautilus-lab-ns.com/auth","ssl-required":"NONE","bearer-only":false,"public-client":false,"resource":"pravega-controller","confidential-port":0,"credentials":{"secret":"123456"}}"#;
+        let v: KeyCloakJson = serde_json::from_str(json_string).unwrap();
+        assert_eq!(v.realm, "nautilus");
+        assert_eq!(
+            v.auth_server_url,
+            "http://keycloak.jarviscb.nautilus-lab-ns.com/auth"
```
tkaitchuck, line 182: Localhost?

config/src/credentials.rs
```diff
@@ -37,26 +41,152 @@ impl Credentials {
...
 
     pub fn get_request_metadata(&self) -> String {
         format!("{} {}", self.method, self.token)
     }
 }
 
+#[derive(Serialize, Deserialize)]
+struct KeyCloakJson {
+    realm: String,
+    #[serde(rename(deserialize = "bearer-only"))]
```
sarlaccpit, line 86: bearer-only, ssl_required, confidential_port , public_client are not needed to be parsed I think.   You really just need the auth-server-url, credentials and resource(aka client_id)

config/src/credentials.rs
```diff
@@ -37,26 +41,152 @@ impl Credentials {
...
+        "audience": AUDIENCE.to_owned(),
+    });
+
+    let path = base_url.to_owned() + &url.to_owned();
+
+    let mut header_map = HeaderMap::new();
+    let bearer = format!("{} {}", BEARER, token);
+    header_map.insert(AUTHORIZATION, bearer.parse().unwrap());
+    let rpt = send_http_request(&path, payload, header_map).await?;
+    Ok(rpt.access_token)
```
sarlaccpit, line 141: I'm not sure if it's in the scope of this PR, but I just wanted to point out that in the response that contains the token in the "access_token" json field, there's also a "expires_in" field (in seconds).

This can be helpful for some soft caching of the rpt such that you can intelligently decide to not make a request for the rpt (not the access token) if you know you have one and it's far enough away from the expires in time.

(doesn't preclude from having to handle token expiration/errors coming back from the pravega controller when the token is bad; but it can optimize traffic to Keycloak.

config/src/lib.rs
```diff
@@ -101,22 +95,8 @@ impl ClientConfigBuilder {
...
-            .collect::<HashMap<String, String>>();
-        if ret_val.contains_key(AUTH_METHOD) {
-            let method = ret_val.get(AUTH_METHOD).expect("get auth method").to_owned();
-            let token = ret_val.get(AUTH_TOKEN).expect("get auth token").to_owned();
-            Credentials::new(method, token)
-        } else {
-            Credentials::default("admin".into(), "1111_aaaa".into())
-        }
+    fn default_credentials(&self) -> Credentials {
+        Credentials::basic("admin".into(), "1111_aaaa".into())
```
tkaitchuck, line 99: Empy string?

tkaitchuck (approved):

config/src/credentials.rs
Tristan1900: Thanks for the comment! 

My understanding is that once an instance of Rust client gets the rpt token, it will never ask for it again. Is that correct?
If that's the case, I think we don't need cache here because rust keycloak client will never be called twice.

I am guessing that the reason that java keycloak client has cache is perhaps Java keycloak client is a service and every Pravega client needs to call it?

sarlaccpit: Just to clarify a couple things:

I don't know how long the Rust client lives; and if perhaps it's interaction with Pravega is a one shot deal.  But just in case it's not; all I meant is that the RPT token that gets a hold of will expire after 5 minutes.

The PravegaKeycloakCredentials is not a service per say; it's merely just a library, implementation of the Pravega client side's Credentials interface which gets loaded at runtime.   The reason why it caches the RPT for some time, is that it knows it might live in the context of a long running application where the Pravega Client remains loaded for a long time.  As such, it needs to be able to get new tokens when asked; and if possible be a bit smart about it a cache the said token.

I don't know if all of this translate to this use case and how the rust client is actually used.   Does it make sense?

sarlaccpit (approved):
My line of comments on the caching are more on the side of optimization.   But it's not critical to this PR.

Meaning, once you have obtained the RPT; and it's sent to the controller, exchanged for a pravega delegation token; the  client process can hold on to that token for some time and reuse it while the delegation token is good.

Only after that will you be forced to reauthenticate.

PR #209, opened 2020-12-17
Issue 198: ConditionalCheckFailed due to reconnection

### LINKED ISSUE #198: conditional check failed during byte stream write
**Problem description**
Running a Pravega write benchmark returned the error `conditional check failed` after 5 minutes.
```
[2020-12-03T19:55:48Z DEBUG pravega_client_rust::reactor::segment_writer] setting up append for writer:WriterId(199441243869397512482901218303664300474)/segment:ScopedSegmen
t { scope: Scope { name: "examples" }, stream: Stream { name: "pbr-15-100000MB-1seg-pod0" }, segment: Segment { number: 0, tx_id: None } } 
[2020-12-03T19:55:48Z DEBUG pravega_client_rust::reactor::segment_writer] append setup completed for writer:WriterId(199441243869397512482901218303664300474)/segment:ScopedS
egment { scope: Scope { name: "examples" }, stream: Stream { name: "pbr-15-100000MB-1seg-pod0" }, segment: Segment { number: 0, tx_id: None } } 
[2020-12-03T19:55:48Z INFO  pravega_client_rust::reactor::segment_writer] finished setting up connection 
[2020-12-03T19:55:48Z DEBUG pravega_client_rust::reactor::segment_writer] flushing 1 events of total size 5242880 to segment "examples/pbr-15-100000MB-1seg-pod0/0.#epoch.0" 
based on offset Some(20683161600); event segment writer id WriterId(199441243869397512482901218303664300474)/connection id: 42a66895-5739-498f-9da7-020504476b62 
[2020-12-03T19:55:48Z WARN  pravega_client_rust::reactor::reactors] conditional check failed ConditionalCheckFailedCommand { writer_id: 199441243869397512482901218303664300474, event_number: 3946, request_id: 3949 }
```

The benchmark application used is https://eos2git.cec.lab.emc.com/faheyc/historical-stream-processing/tree/master/pravega-benchmark-rust. This uses pravega-client-rust commit 2c03fd8. The server uses https://github.com/claudiofahey/pravega/commit/101f0894b77e2d0bcf95eacf77f44647d61bbc9a.

**Problem location**
Unknown.

Unknown.

Tristan1900:
**Change log description**  

A false `ConditionalCheckFailed` returned due to reconnection. 

What happens is that when client sends an Event `e` to server and server has already accepted `e`, a reconnection is triggered before client receives the ack reply. The `e` is sent to server again due to reconnection so the ConditionalCheckFailed is returned since `e` has already been accepted before.

The temporary fix for this issue it to check if a reconnection happened right before receiving `ConditionalCheckFailed`. If it is then this error could be a false alarm and we can ignore it. However, this is not 100% guaranteed since another writer could write some data to the segment right at the reconnection session of our writer, and in that case the `ConditionalCheckFailed` is true. Right now there is no good way to tell unless we check the data in segment. But based on the discussion in #198, the temporary fix is probably good enough.

**Purpose of the change**  
Fix #198 

NOTE: more investigations are needed, please ignore it for now

### DIFF
--- a/src/reactor/reactors.rs
+++ b/src/reactor/reactors.rs
@@ -18,6 +18,8 @@ use crate::client_factory::ClientFactory;
 use crate::reactor::event::{Incoming, ServerReply};
 use crate::reactor::segment_selector::SegmentSelector;
 
+const MAX_RECONNECTION_ALLOWED_FOR_CONDITIONAL_CHECK: i32 = 3;
+
 #[derive(new)]
 pub(crate) struct Reactor {}
 
@@ -110,8 +112,8 @@ impl Reactor {
                     "data appended for writer {:?}, latest event id is: {:?}",
                     writer.id, cmd.event_number
                 );
-                // reconnection works as expected, set this flag to false.
-                writer.reconnect = false;
+                // reconnection works as expected, set reconnection to 0
+                writer.reconnection = 0;
                 writer.ack(cmd.event_number);
                 if let Err(e) = writer.write_pending_events().await {
                     warn!(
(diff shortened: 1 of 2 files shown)


src/reactor/reactors.rs
```diff
@@ -162,9 +164,23 @@ impl Reactor {
...
+                    if writer.reconnect {
+                        // Note: it still could be caused by interleaved data, but the probability is low.
+                        // Right now there is no good way to tackle this issue. Only application side
+                        // can tell if this is a false alarm or not.
+                        warn!(
+                            "conditional check failed {:?}, probably a false alarm caused by reconnection",
+                            cmd
+                        );
+                        writer.ack(cmd.event_number);
+                        writer.reconnect = false;
```
tkaitchuck, line 177: Based on the discussion here:
https://github.com/pravega/pravega-client-rust/issues/198#issuecomment-747818033
It seems like the thing to do is that if `reconnect` is `true` then treat `ConditionalCheckFailed` as a normal connection failure and trigger another re-connect.

tkaitchuck (approved):

PR #212, opened 2021-01-06
Perf improvement: Avoid a copy of the routing_key string on every write.

tkaitchuck:
Signed-off-by: Tom Kaitchuck <tom.kaitchuck@emc.com>

**Change log description**  
 Avoid a copy of the routing_key string on every write.

**What the code does**  
Borrows the ScopedSegment rather than performing a clone of the string.


Tristan1900 (approved):
Great! Thanks!

PR #218, opened 2021-01-11
Issue 217: Table map uses wrong scope

### LINKED ISSUE #217: Wrong scope in tablemap
**Problem description**
Tablemap is using `_table` scope from the first commit. It's not revealed to be a bug until we restrict the permission of the client. We probably want to use something like `<user scope>/_table+name/0` as the internal segment.

Update: segment name as `<user scope>/<table name>_kvtable/0`

**Problem location**
tablemap

Tristan1900:
**Change log description**  
Table map should use segment under the a user configured scope

**Purpose of the change**  
Fix #217 

**What the code does**  
Let user pass a scope when creating the table map.

**How to verify it**  
Test should pass

### DIFF
--- a/bindings/Cargo.toml
+++ b/bindings/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pravega"
-version = "0.0.1-alpha-3"
+version = "0.0.1-alpha-4"
 edition = "2018"
 categories = ["Network programming"]
 keywords = ["streaming", "client", "pravega"]
@@ -11,7 +11,7 @@ description = "Pravega client"
 authors = ["Pravega Community"]
 
 [package.metadata.maturin]
-classifier = ["Development Status :: 3 - Alpha", "License :: OSI Approved :: Apache Software License",
+classifier = ["Development Status :: 4 - Alpha", "License :: OSI Approved :: Apache Software License",
     "Programming Language :: Rust", "Programming Language :: Python :: 3.8", "Programming Language :: Python :: 3.9",]
 requires-python = ">= 3.8"
 project-url = ["Homepage, https://pravega.github.io/pravega-client-rust/"]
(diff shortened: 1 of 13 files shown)


bindings/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pravega"
-version = "0.0.1-alpha-3"
+version = "0.0.1-alpha-4"
```
shrids, line 3: The change will be published only in case we tag a branch ref: https://github.com/pravega/pravega-client-rust/blob/7a0cd9d217bd742824ff209a5c4d9c09847c12c2/.github/workflows/tagPublish.yml#L6 
This can be changed...

shrids (approved):

src/client_factory.rs
```diff
@@ -113,17 +113,19 @@ impl ClientFactory {
             "Creating reader group {:?} to read data from stream {:?}",
             reader_group_name, stream
         );
+        let scope = stream.scope.clone();
         let rg_config = ReaderGroupConfigBuilder::default().add_stream(stream).build();
-        ReaderGroup::create(reader_group_name, rg_config, self.clone()).await
+        ReaderGroup::create(scope, reader_group_name, rg_config, self.clone()).await
```
tkaitchuck, line 118: This logic is wrong. A reader group can read from many streams and they can even be in different scopes. So the scope of the reader group should NOT be the same as the scope for the stream.

tkaitchuck (changes requested):

bindings/Cargo.toml
Tristan1900: Thanks. I increment it since some change has been made, so if we build it from scratch it would be different than the published one. Right now I'm not trying to publish it, only to build locally and test it.

tkaitchuck (approved):

PR #229, opened 2021-01-27
Issue 228: Fix acquire segment

### LINKED ISSUE #228: Multiple readers cannot work properly
**Problem description**
The observation is that, if we have 2 readers A and B, and they read a stream of 2 segments. Reader A starts first and it will acquire all segments, B starts after A and acquires zero segment. After a while reader A release one of the segments due to rebalancing, however reader B cannot get that segment and that segment is left unread. 

**Problem location**
The problem is [here](https://github.com/pravega/pravega-client-rust/blob/master/src/event_reader.rs#L556), reader B gets stuck at this line and never get a chance to return from `acquire_segment` method. So it cannot re-enter the method and acquire any available segments.

use `try_recv` instead of `recv`

Tristan1900:
**Change log description**  
Change `recv` to `try_recv` in `acquire_segment` method.

**Purpose of the change**  
Fix #228 

**What the code does**  
`recv` will block forever if there is no slice for that reader. Use `try_recv` to let method return and caller can re-enter the method to acquire new segments.

**How to verify it**  
Test should pass

### DIFF
--- a/integration_test/src/event_stream_reader_tests.rs
+++ b/integration_test/src/event_stream_reader_tests.rs
@@ -102,7 +102,7 @@ fn test_multi_reader_multi_segments_tail_read(client_factory: &ClientFactory, rt
     let scope_name = Scope::from("testMultiReaderMultiSegmentsTailRead".to_owned());
     let stream_name = Stream::from("testMultiReaderMultiSegmentsTailRead".to_owned());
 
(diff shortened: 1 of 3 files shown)


tkaitchuck (approved):

shrids (approved):
LGTM. 
(I did not get a chance to verify it locally...)

PR #235, opened 2021-03-05
Fix release error

Tristan1900:
**Change log description**  
Fixes errors when doing `cargo publish`


tkaitchuck (approved):

Cargo.toml
```diff
@@ -38,14 +38,14 @@ members = [
...
 [dependencies]
-pravega-client-shared = { path = "./shared"}
-pravega-controller-client = { path = "./controller-client"}
-pravega-wire-protocol = { path = "./wire_protocol"}
-pravega-client-retry = {path = "./retry"}
-pravega-connection-pool = {path = "connection_pool" }
-pravega-client-channel = {path = "./channel"}
-pravega-client-auth = {path = "./auth"}
-pravega-client-config = {path = "./config"}
+pravega-client-shared = "0.1"
```
shrids, line 41: Is there an option to put 0.1 in a variable? ; this would simplify work for future releases...

shrids (approved):
No blocking comments from my end here...

Cargo.toml
Tristan1900: That's a good suggestion but I don't think Cargo.toml file allows variable. It appears that they ruled out such suggestions.

bindings/Cargo.toml
```diff
@@ -14,7 +14,7 @@ authors = ["Pravega Community"]
 classifier = ["Development Status :: 5 - Production/Stable", "License :: OSI Approved :: Apache Software License",
     "Programming Language :: Rust", "Programming Language :: Python :: 3.8", "Programming Language :: Python :: 3.9",]
 requires-python = ">= 3.8"
-project-url = ["homepage = https://pravega.github.io/pravega-client-rust/"]
+project-url = {"homepage" = "https://pravega.github.io/pravega-client-rust/"}
```
shrids, line 17: This change prevents the working of maturin version configured in the workflow

Tristan1900: Nice catch! Some CI is using 0.8.3 while others are using 0.9.0+.

.github/workflows/tagPublish.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           components: rustfmt, clippy
       - run: rustup set default-host ${{ matrix.platform.rust-target }}
       - name: install maturin
-        run: pip install 'maturin==0.8.3'
```
shrids, line 38: Could you confirm maturin 0.9.0 allows `1-unchecked` option used `build_wheel.py`  file?  With the latest maturin version, I had obeserved that it does not work. 
Note: we cannot upload python bindings to pypi repo if manylinux option is turned off and we have used 1-unchecked since auth so libraries used **"libssl.so.1.1", "libcrypto.so.1.1"** is not manylinux compliant.

Tristan1900: Thanks Sandeep. You are right that we cannot upgrade to 0.9.0 as it does not support `1-unchecked`.

PR #242, opened 2021-04-14
Issue 225: Implement list streams controller API

### LINKED ISSUE #225: Implement list_streams API
**Problem description**
Controller list_streams is not implemented.

**Problem location**
https://github.com/pravega/pravega-client-rust/blob/master/controller-client/src/lib.rs#L328

None.

shrids:
**Change log description**  
- Enable List Streams Controller API on the Rust Client.
- Create a Paginator utility which returns a `impl Stream<Item = Result<String, RetryError<ControllerError>>>`  to handle huge stream lists with pagination.
- Enable List Streams option in the Controller CLI

**Purpose of the change**  
Fixes #225 

**What the code does**  
- Enable controller api list streams which returns the streams for the provided scope and continuationToken.
```rust
async fn list_streams(&self, scope: &Scope, token: &str) -> ResultRetry<Option<(Vec<String>, String)>>;
```
- A paginator helper which returns an `impl Stream`  thereby lazily fetch the next page of streams from Pravega
```rust
pub fn list_streams(
    scope: Scope,
    client: &dyn ControllerClient,
) -> impl Stream<Item = Result<String, RetryError<ControllerError>>> 
```
This uses the `futures::stream::unfold` to fetch streams under a given scope.
- Example usage 
```rust
// Sample 1
use pravega_client_shared::Scope;
use pravega_controller_client::paginator::list_streams;
    let stream = list_streams(
        Scope {
            name: "testScope".to_string(),
        },
        controller_client,
    );
    // collect all the Streams in a single vector
    let stream_list:Vec<String> = stream.map(|str| str.unwrap()).collect::<Vec<String>>().await;

// Sample 2
use pravega_client_shared::Scope;
use pravega_controller_client::paginator::list_streams;
    let mut stream = list_streams(
        Scope {
             name: "testScope".to_string(),
         },
        controller_client,
    );
let pravega_stream_1 = stream.next().await;
let pravega_stream_2 = stream.next().await;
// A None is returned at the end of the stream.
```

CLI Option for list-streams.
```
osboxes@osboxes:~/$ ./controller-cli list-streams --help
controller-cli-list-streams 0.1.0
List Streams under a scope

USAGE:
    controller-cli list-streams <scope-name>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

ARGS:
    <scope-name>    Scope Name

```

**How to verify it**  
The existing and newly added tests should pass. 
Verified it for a huge list of streams against standalone.

### DIFF
--- a/controller-client/src/lib.rs
+++ b/controller-client/src/lib.rs
@@ -26,7 +26,6 @@
 #![allow(clippy::multiple_crate_versions)]
 #![allow(dead_code)]
 #![allow(clippy::similar_names)]
-#![allow(clippy::upper_case_acronyms)]
 
 use std::result::Result as StdResult;
 use std::time::{Duration, Instant};
--- a/controller-client/src/lib.rs
+++ b/controller-client/src/lib.rs
@@ -1542,9 +1542,9 @@ mod test {
 
         // test list streams
         let res = rt
-            .block_on(controller.list_streams(&scope))
+            .block_on(controller.list_streams(&scope, &String::from("")))
             .expect("list streams");
-        assert_eq!(res, vec!["s1".to_string()])
+        assert_eq!(res, Some((vec!["s1".to_string()], String::from("123"))))
     }
 
     #[derive(Default)]
--- a/controller-client/src/lib.rs
+++ b/controller-client/src/lib.rs
@@ -133,7 +133,11 @@ pub trait ControllerClient: Send + Sync {
      * API to list streams under a given scope and continuation token.
      * Use the pravega_controller_client::paginator::list_streams to paginate over all the streams.
      */
-    async fn list_streams(&self, scope: &Scope, token: &str) -> ResultRetry<Option<(Vec<String>, String)>>;
+    async fn list_streams(
+        &self,
+        scope: &Scope,
+        token: &CToken,
+    ) -> ResultRetry<Option<(Vec<ScopedStream>, CToken)>>;
 
     /**
      * API to delete a scope. Note that a scope can only be deleted in the case is it empty. If
@@ -331,7 +335,11 @@ impl ControllerClient for ControllerClientImpl {
         )
     }
 
-    async fn list_streams(&self, scope: &Scope, token: &str) -> ResultRetry<Option<(Vec<String>, String)>> {
+    async fn list_streams(
+        &self,
+        scope: &Scope,
+        token: &CToken,
+    ) -> ResultRetry<Option<(Vec<ScopedStream>, CToken)>> {
         wrap_with_async_retry!(
             self.config.retry_policy.max_tries(MAX_RETRIES),
             self.call_list_streams(scope, token)
@@ -612,13 +620,15 @@ impl ControllerClientImpl {
         })
     }
 
-    async fn call_list_streams(&self, scope: &Scope, token: &str) -> Result<Option<(Vec<String>, String)>> {
+    async fn call_list_streams(
+        &self,
+        scope: &Scope,
+        token: &CToken,
+    ) -> Result<Option<(Vec<ScopedStream>, CToken)>> {
         let operation_name = "ListStreams";
         let request: StreamsInScopeRequest = StreamsInScopeRequest {
             scope: Some(ScopeInfo::from(scope)),
-            continuation_token: Some(ContinuationToken {
-                token: token.to_string(),
-            }),
+            continuation_token: Some(ContinuationToken::from(token)),
         };
         debug!(
             "Triggering a request to the controller to list streams for scope {}",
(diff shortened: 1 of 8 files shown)


Tristan1900:
Thanks Sandeep! That's really nice!
One thing that is a little bit odd is that `list_stream` is a controller method but it's not used as `controller.list_stream()` like other methods. Could we still use `controller.list_stream()` and let it return the `impl stream`?

shrids:
Yes @Tristan1900 , that was my initial plan. I later realized that we cannot have `impl Stream`  inside the trait definition.
i.e. 
```rust
pub trait ControllerClient: Send + Sync {
    fn list_stream(&self, scope: &Scope) -> impl Stream<Item = ResultRetry<String>> // this is not allowed.
}
```

Tristan1900:
@shrids Thanks yeah I didn't know that. Looks like `impl trait` is not allowed to be in return type. But can we do `Box<Stream>`?

controller-client/src/lib.rs
```diff
@@ -604,6 +612,57 @@ impl ControllerClientImpl {
...
+
+                    let token: Option<ContinuationToken> = result.continuation_token;
+                    let ct: String = match token.map(|t| t.token) {
+                        None => {
+                            warn!(
+                                "None returned for continuation token list streams API for scope {}",
+                                scope
+                            );
+                            String::from("")
+                        }
```
tkaitchuck, line 650: Doesn't this just mean there are no more items? Why the warning? and why the empty string?

controller-client/src/paginator.rs
```diff
@@ -0,0 +1,121 @@
...
+use std::vec::IntoIter;
+use tracing::info;
+
+///
+///Helper method to iterated over the all the Pravega streams under the provided Scope.
+///This method returns a stream of values,Pravega streams, produced asynchronously.
+///
+/// The below snippets show case the example uses.
+/// Sample 1:
+///```ignore
```
tkaitchuck, line 25: Do you want `ignore` or `no_run`?

controller-client/src/paginator.rs
```diff
@@ -0,0 +1,121 @@
...
+///         },
+///         controller_client,
+///     );
+///     // collect all the Streams in a single vector
+///     let stream_list:Vec<String> = stream.map(|str| str.unwrap()).collect::<Vec<String>>().await;
+///  # });
+/// ```
+///
+/// Sample 2:
+/// ```ignore
```
tkaitchuck, line 49: same

controller-client/src/paginator.rs
```diff
@@ -0,0 +1,121 @@
...
+/// let pravega_stream_1 = stream.next().await;
+/// let pravega_stream_2 = stream.next().await;
+/// // A None is returned at the end of the stream.
+///  # });
+/// ```
+///
+pub fn list_streams(
+    scope: Scope,
+    client: &dyn ControllerClient,
+) -> impl Stream<Item = Result<String, RetryError<ControllerError>>> + '_ {
```
tkaitchuck, line 77: Should this be a `Stream` rather than a string?

controller-client/src/paginator.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    // Initial state with an empty Continuation token.
+    stream::unfold(
+        State {
+            streams: Vec::new().into_iter(),
+            scope,
+            token: String::from(""),
+        },
+        move |mut state| async move {
```
tkaitchuck, line 91: Make this a private function and pass the method here, rather than creating an anonymous one. It will look cleaner.

tkaitchuck (changes requested):

controller-client/src/lib.rs
```diff
@@ -128,7 +129,11 @@ pub trait ControllerClient: Send + Sync {
      */
     async fn create_scope(&self, scope: &Scope) -> ResultRetry<bool>;
 
-    async fn list_streams(&self, scope: &Scope) -> ResultRetry<Vec<String>>;
+    /**
+     * API to list streams under a given scope and continuation token.
+     * Use the pravega_controller_client::paginator::list_streams to paginate over all the streams.
+     */
+    async fn list_streams(&self, scope: &Scope, token: &str) -> ResultRetry<Option<(Vec<String>, String)>>;
```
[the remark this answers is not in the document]
shrids: fixed it.

controller-client/src/paginator.rs
shrids: changed it to ScopedStream

controller-client/src/paginator.rs
shrids: i used `ignore` since

controller-client/src/paginator.rs
shrids: improved it.

controller-client/src/lib.rs
shrids: improved the code behaviour here.

controller-client/src/paginator.rs
tkaitchuck: Oh, yeah that a different crate. Sorry I don't think there is a way around that which doesn't involve adding a dev-dependency.

tkaitchuck (approved):

PR #251, opened 2021-04-29
Issue 245: Async token refresh

### LINKED ISSUE #245: Keycloak authentication token does not get refreshed
**Problem description**
An application that uses the byte stream reader to read from Pravega with Keycloak authentication works well for 5 minutes and then it panics. This is likely caused by not refreshing the token.

```
thread 'tokio-runtime-worker' panicked at 'get endpoint for async segment reader: RetryError { error: OperationError { can_retry: false, operation: "get_endpoint", error_msg: "status: Unauthenticated, message: \"\", details: [], metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"authorization\": \"Bearer snipped\"} }" }, total_delay: 0ns, tries: 1 }', /mnt/home/faheyc/nautilus/gstreamer/gstreamer-pravega/pravega-client-rust/src/segment_reader.rs:229:30
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/panicking.rs:92:14
   2: core::option::expect_none_failed
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/option.rs:1300:5
   3: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   4: <tokio::future::poll_fn::PollFn<F> as core::future::future::Future>::poll
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   6: std::panicking::try
   7: tokio::runtime::task::harness::poll_future
   8: tokio::runtime::task::harness::Harness<T,S>::poll
   9: std::thread::local::LocalKey<T>::with
  10: tokio::runtime::thread_pool::worker::Context::run_task
  11: tokio::runtime::thread_pool::worker::Context::run
  12: tokio::macros::scoped_tls::ScopedKey<T>::set
  13: tokio::runtime::thread_pool::worker::run
  14: tokio::loom::std::unsafe_cell::UnsafeCell<T>::with_mut
  15: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  16: tokio::runtime::task::harness::Harness<T,S>::poll
  17: tokio::runtime::blocking::pool::Inner::run

```

**Problem location**
Keycloak client.

None.

Tristan1900:
**Change log description**  
Currently token refresh is using a Runtime which makes it impossible to work in a async context. Use async way to refresh the token instead.

**Purpose of the change**  
Fix #245 

**What the code does**  
Use async way to refresh token

### DIFF
--- a/config/src/credentials.rs
+++ b/config/src/credentials.rs
@@ -72,6 +72,10 @@ impl Credentials {
     pub async fn get_request_metadata(&self) -> String {
         self.inner.get_request_metadata().await
     }
+
+    pub fn is_expired(&self) -> bool {
+        self.inner.is_expired()
+    }
 }
 
 impl Clone for Credentials {
@@ -85,6 +89,7 @@ impl Clone for Credentials {
 #[async_trait]
 trait Cred: Debug + CredClone + Send + Sync {
     async fn get_request_metadata(&self) -> String;
+    fn is_expired(&self) -> bool;
 }
 
 trait CredClone {
@@ -111,6 +116,10 @@ impl Cred for Basic {
     async fn get_request_metadata(&self) -> String {
         format!("{} {}", self.method, self.token)
     }
+
+    fn is_expired(&self) -> bool {
+        false
+    }
 }
 
 #[derive(Debug, Clone)]
@@ -128,6 +137,13 @@ impl Cred for KeyCloak {
             self.refresh_rpt_token().await;
         }
         format!("{} {}", self.method, *self.token.lock().await)
+    }
+
+    fn is_expired(&self) -> bool {
+        let now = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("get unix time");
+        now.as_secs() + REFRESH_THRESHOLD_SECONDS >= self.expires_at.load(Ordering::Relaxed)
     }
 }
 
@@ -168,13 +184,6 @@ impl KeyCloak {
 
         *self.token.lock().await = rpt.access_token;
         self.expires_at.store(expires_at, Ordering::Relaxed);
-    }
-
-    fn is_expired(&self) -> bool {
-        let now = SystemTime::now()
-            .duration_since(SystemTime::UNIX_EPOCH)
-            .expect("get unix time");
-        now.as_secs() + REFRESH_THRESHOLD_SECONDS >= self.expires_at.load(Ordering::Relaxed)
     }
 }
 
--- a/config/src/credentials.rs
+++ b/config/src/credentials.rs
@@ -16,9 +16,8 @@ use std::fs::File;
 use std::io::{BufReader, Read};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
-use std::sync::Mutex;
 use std::time::SystemTime;
-use tokio::runtime::Runtime;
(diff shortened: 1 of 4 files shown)


config/src/credentials.rs
```diff
@@ -124,17 +123,16 @@ struct KeyCloak {
 
 #[async_trait]
 impl Cred for KeyCloak {
-    fn get_request_metadata(&self) -> String {
+    async fn get_request_metadata(&self) -> String {
         if self.is_expired() {
-            self.refresh_rpt_token();
+            self.refresh_rpt_token().await;
```
shrids, line 128: A token refresh will be initiated only when there are about 5 seconds for the token to expire. But this method is not invoked automatically every time a request is made. Shouldn't it be the case?

controller-client/src/lib.rs
```diff
@@ -556,12 +556,28 @@ impl ControllerClientImpl {
     /// which runs the connection in a background task and provides a `mpsc` channel interface.
     /// Due to this cloning the `Channel` type is cheap and encouraged.
     ///
-    fn get_controller_client(&self) -> ControllerServiceClient<Channel> {
-        self.channel.read().unwrap().clone()
+    async fn get_controller_client(&self) -> ControllerServiceClient<Channel> {
+        if self.config.is_auth_enabled && self.config.credentials.is_expired() {
+            // get_request_metadata internally checks if token expired before sending request to the server,
+            // race condition might happen here but eventually only one request will be sent.
+            let mut x = self.channel.write().await;
```
shrids, line 563: we should remove code duplication of this logic here and `reset()`.

shrids (approved):
Apart from the code duplication comment the changes look good.

controller-client/src/lib.rs
Tristan1900: Thanks!

Tristan1900: Just realize that we cannot remove the logic in reset. The reason is that `basic` method will never expire so if we remove the logic in reset then after resetting the auth header will be gone.

wire_protocol/src/connection.rs
```diff
@@ -175,6 +175,15 @@ impl Connection for TlsConnection {
...
             .write_all(payload)
             .await
+            .context(SendData {
+                endpoint: endpoint.clone(),
+            })?;
+
+        self.stream
+            .as_mut()
+            .expect("get connection")
+            .flush()
```
shrids, line 185: This API Flushes this stream, ensuring that all intermediately buffered contents reach their destination.
https://docs.rs/tokio-rustls/0.22.0/src/tokio_rustls/client.rs.html#93-94 snippet indicates a `TlsStream.flush()` is required to ensure data actually reaches the destination. 
Is this different with a TcpStream ?

Tristan1900: yeah I think TcpStream itself doesn't have a buffer and any write is directly sent over the wire. However this TlsStream has a buffer inside that needs to be flushed. The name `TlsStream` is kind of misleading.

Issue #257, opened 2021-05-12
Unable to manually mark a stopped reader as offline


Mikey032:
**Problem description**
When you don't explicitly close a reader it will be stuck forever in an online state because it's never detected as failed and there are no methods to close it without creating a reader.

Sample code:
``` python
reader_group = manager.create_reader_group(readergroup, scope, stream)
reader = reader_group.create_reader(reader)
```
thread '<unnamed>' panicked at 'Error while creating the reader: SyncError { error_msg: "add reader Reader { name: \"<reader>\" }", source: SyncUpdateError { error_msg: "Failed to add online reader Reader { name: \"reader\" }: reader already online" } }

**Problem location**
python client binding

**Suggestions for an improvement**
Implement methods to list all readers in a reader group and be able to close them by name/id.

tkaitchuck:
This can be done but it currently involves manually creating the object, which obviously is a terrible UX.
We can add a method on the Group to do this.

Mikey032:
That would be great👍

PR #261, opened 2021-05-14
Issue 200: Drop on EventReader should automatically invoke reader offline API.

### LINKED ISSUE #200: Drop on EventReader should invoke readeroffline
**Problem description**

- Drop on an EventReader should invoke reader offline.

**Problem location**
Reader.

Note that TableSynchronizer/ReaderGroupState.remove_reader throws an error if the reader is already offline.

shrids:
**Change log description**  
Drop on EventReader should invoke readerOffline.

**Purpose of the change**  
Fixes #200 

**What the code does**  
Drop on EventReader should invoke `reader_offline()`.

This is useful in scenarios where the Python user forgets to invoke reader offline. A scenario described in #257 is one instance of this.

With this change, `reader_offline()` is automatically invoked when the object is destructed.
```
In [11]: r1=rg.create_reader("r1")
May 14 18:54:41.627  INFO pravega_client::stream_reader_group: Creating reader "r1" under reader group "rg2"
May 14 18:54:41.627  INFO pravega_client::event::reader_group_state: Adding reader Reader { name: "r1" } to reader group

In [12]: r1=rg.create_reader("r2")
May 14 18:54:47.361  INFO pravega_client::stream_reader_group: Creating reader "r2" under reader group "rg2"
May 14 18:54:47.361  INFO pravega_client::event::reader_group_state: Adding reader Reader { name: "r2" } to reader group
May 14 18:54:47.369  INFO pravega_client::event::reader: putting reader r1 offline
```

**How to verify it**  
All the existing tests should continue to pass.

### DIFF
--- a/src/event/reader.rs
+++ b/src/event/reader.rs
@@ -23,12 +23,14 @@ use im::HashMap as ImHashMap;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
+use tokio::runtime::Handle;
 use tokio::sync::mpsc::{Receiver, Sender};
 use tokio::sync::oneshot;
 use tokio::sync::oneshot::error::TryRecvError;
 use tokio::sync::{mpsc, Mutex};
 use tokio::time::timeout;
 use tracing::{debug, error, info, warn};
+
 type ReaderErrorWithOffset = (ReaderError, i64);
 type SegmentReadResult = Result<SegmentDataBuffer, ReaderErrorWithOffset>;
 
@@ -111,8 +113,22 @@ impl Drop for EventReader {
     /// Destructor for reader invoked. This will automatically invoke reader_offline().
     fn drop(&mut self) {
         info!("reader {} is dropped", self.id);
-        let runtime = self.factory.runtime().handle().clone();
-        runtime.block_on(self.reader_offline());
+        // try fetching the currently running Runtme.
+        let r = Handle::try_current();
+        match r {
+            Ok(handle) => {
+                // enter the runtime context.
+                let _ = handle.enter();
+                // ensure we block until the reader_offline method completes.
+                futures::executor::block_on(self.reader_offline());
+            }
+            Err(_) => {
+                // get a handle of the runtime.
+                let handle = self.factory.runtime().handle().clone();
+                // ensure we block until the reader_offline executes successfully.
+                handle.block_on(self.reader_offline());
+            }
+        }
     }
 }
 
@@ -218,6 +234,7 @@ impl EventReader {
                 slice_stop_reading,
                 last_segment_release: Instant::now(),
                 last_segment_acquire: Instant::now(),
+                reader_offline: false,
             },
             rg_state,
         }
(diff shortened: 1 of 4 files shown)


tkaitchuck (approved):
Please add a test.

src/event/reader.rs
```diff
@@ -107,6 +107,15 @@ pub struct EventReader {
     rg_state: Arc<Mutex<ReaderGroupState>>,
 }
 
+impl Drop for EventReader {
+    /// Destructor for reader invoked. This will automatically invoke reader_offline().
+    fn drop(&mut self) {
+        info!("reader {} is dropped", self.id);
+        let runtime = self.factory.runtime().handle().clone();
+        runtime.block_on(self.reader_offline());
```
Tristan1900, line 115: Yeah I don't think we can call `block_on` in `Drop`, this will prevent `EventReader` from being used in async contexts.
I once tried to use channel to `blocking_send` a message in `drop` and it panicked.

Tristan1900 (changes requested):

src/event/reader.rs
Tristan1900: This is tricky as I couldn't come up with a way to use async function in `Drop`. Perhaps eventually we still need to ask user to call offline manually?

tkaitchuck: You cannot `,await` inside of a drop.

tkaitchuck: I think this problem is causing the build failure.

tkaitchuck: @Tristan1900 found this: https://docs.rs/tokio/1.6.0/tokio/task/index.html#block_in_place
which sounds like the solution we are looking for.

Tristan1900: probably we want to use this 
https://docs.rs/tokio/1.6.0/tokio/task/index.html#block_in_place

tkaitchuck: Based on what we talked about in the meeting:
The part of `reader_offline` that needs to be async is:
```
        self.rg_state
            .lock()
            .await
            .remove_reader(&self.id, offset_map)
            .await
            .expect("Update ReaderGroup to ensure reader is offline");
```
I think we can factor this out into a function which `reader_offline` can also call. Then we can do the sync work here (maybe also a private function up to you) and pass the map to the async method inside a `spawn`. The map can be moved, and the `rg_state` is an arc, so it can just be cloned.

shrids: I have updated the code with the solution.

src/event/reader.rs
```diff
@@ -107,6 +109,29 @@ pub struct EventReader {
...
+    fn drop(&mut self) {
+        info!("reader {} is dropped", self.id);
+        // try fetching the currently running Runtme.
+        let r = Handle::try_current();
+        match r {
+            Ok(handle) => {
+                // enter the runtime context.
+                let _ = handle.enter();
+                // ensure we block until the reader_offline method completes.
+                futures::executor::block_on(self.reader_offline());
```
Tristan1900, line 123: I check around and see people are saying that this way is not encouraged. I guess we use this workaround for now and come back if better solution comes up.

Tristan1900 (approved):

PR #263, opened 2021-05-17
Issue 262: support multiple segments in byte stream

### LINKED ISSUE #262: Support multiple segments in byte stream
**Problem description**
ByteStream right now only contains 1 segment. We should support multiple segments in the byte stream. 
The prerequisite is that the stream is configured as fixed size of segments.

Tristan1900:
**Change log description**  
support multiple segments in byte stream

**Purpose of the change**  
Fix #262 

**What the code does**  
Add a `Option<ScopedSegment>` field in the `PendingEvent`. 
If the `Option` is not `None`, then `selector` will directly fetch the writer from the map using `segment` as a key.

**How to verify it**  
Tests should pass

### DIFF
--- a/src/segment/event.rs
+++ b/src/segment/event.rs
@@ -38,9 +38,14 @@ pub(crate) struct WriterInfo {
 }
 
 #[derive(Debug)]
+pub(crate) enum RoutingInfo {
+    RoutingKey(Option<String>),
+    Segment(ScopedSegment),
+}
+
+#[derive(Debug)]
 pub(crate) struct PendingEvent {
-    pub(crate) scoped_segment: Option<ScopedSegment>,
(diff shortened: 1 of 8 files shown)


src/segment/event.rs
```diff
@@ -39,6 +39,7 @@ pub(crate) struct WriterInfo {
 
 #[derive(Debug)]
 pub(crate) struct PendingEvent {
+    pub(crate) scoped_segment: Option<ScopedSegment>,
     pub(crate) routing_key: Option<String>,
```
tkaitchuck, line 43: Rather than have two options which are mutually exclusive, you could have an enum which has either a segment or a routing key.

tkaitchuck (approved):

Issue #244, opened 2021-04-21
Client should accept multiple CA certificates


claudiofahey:
**Problem description**
The Pravega client currently requires that the caller provide a single CA certificate for validating the connection to the server. However, it is common for a client user to not know exactly which CA certificate to use. In most TLS client implementations, the client will automatically select from a set of multiple CA certificates, based on information provided by the server.

**Problem location**
TLS

**Suggestions for an improvement**
Allow the user to provide a CA bundle or a directory containing multiple CA certs. Alternatively, use OpenSSL which should use the OS trust store.

claudiofahey:
It is preferred that the Pravega Rust Client use the same certs that the OS uses. On Ubuntu, it should use the results of the `update-ca-certs` command.
Additionally, we should use this opportunity to ensure that TLS 1.3 is supported. It is unknown whether TLS 1.3 is currently supported.

PR #265, opened 2021-05-19
Issue 243, 244: Support multiple certs and allow user to pass keycloak as json string

### LINKED ISSUE #243: Authentication parameters should be accepted via the API
**Problem description**
The Pravega client uses environment variables for authentication parameters. This works well for the usual case of connecting to a single Pravega instance. But for a process that copies data from the edge to the cloud, this will not work because different authentication parameters will be needed for each stream.

**Problem location**
Authentication

It should be possible to provide all authentication parameters using the Pravega client API. Allow Keycloak credentials to be provided to the API as a UTF-8 JSON string.

Tristan1900:
**Change log description**  
1. `ClientConfig` can now take multiple certs. Users cam put certs in a directory and pass that directory to the `ClientConfig`.
2. User can now pass keycloak json string directly to the `ClientConfig` by using `Credentials::keycloak_by_json_string(json);`

**Purpose of the change**  
Fix #243 #244 

**What the code does**  
1. Instead of reading from a file to get a cert, we read all the files in a given directory and add all the certs to the config.
2. Have a new method `Credentials::keycloak_by_json_string` that takes the json string as input. Also cache the json string so we don't need to read from files everytime when refreshing. 

**How to verify it**  
Tests should pass


config/src/credentials.rs
```diff
@@ -58,10 +58,33 @@ impl Credentials {
...
-            path: path.to_string(),
+            json: key_cloak_json,
+            expires_at: Arc::new(AtomicU64::new(0)),
+        };
+        Credentials {
+            inner: Box::new(keycloak) as Box<dyn Cred>,
+        }
+    }
+
+    pub fn keycloak_by_json_string(json: &str) -> Self {
```
tkaitchuck, line 81: should that be "from_json_string"?

controller-client/src/lib.rs
```diff
@@ -310,10 +313,33 @@ async fn get_channel(config: &ClientConfig) -> Channel {
...
-        let pem = std::fs::read(&config.trustcert).expect("read truststore");
-        let ca = Certificate::from_pem(pem);
+        let mut rustls_client_config = RustlsClientConfig::new();
+        rustls_client_config.alpn_protocols.push(Vec::from("h2"));
+        let cert_paths =
+            std::fs::read_dir(&config.trustcert).expect("cannot read from the provided cert directory");
+        for entry in cert_paths {
+            let path = entry.expect("get the cert file").path();
+            debug!("reading cert file {}", path.display());
+            let mut pem = BufReader::new(File::open(path.clone()).expect("read cert file"));
```
tkaitchuck, line 323: Two things:
1. This is doing blocking IO in an async method.
2. This is done more than once: `reset` and `refresh_token_if_needed` both call this. - It would be much better if this were done once in the lifetime of the controller-client.

wire_protocol/src/connection_factory.rs
```diff
@@ -99,8 +99,28 @@ impl ConnectionFactory for TokioConnectionFactory {
...
             );
             let mut config = rustls::ClientConfig::new();
-            let mut pem = BufReader::new(File::open(&self.path).expect("open pem file"));
-            config.root_store.add_pem_file(&mut pem).expect("add pem file");
+            let cert_paths =
+                std::fs::read_dir(&self.path).expect("cannot read from the provided cert directory");
+            for entry in cert_paths {
+                let path = entry.expect("get the cert file").path();
+                debug!("reading cert file {}", path.display());
+                let mut pem = BufReader::new(File::open(path.clone()).expect("read cert file"));
```
tkaitchuck, line 107: Same thing here. We should avoid reading files everytime we establish a connection. It would be better to do this once per factory.

tkaitchuck (approved):

shrids (approved):

PR #266, opened 2021-05-24
Issue 264: Manually validate the connection

### LINKED ISSUE #264: Connection should not be recycled if errors happen in the caller.
**Problem description**
Connections should not be put back into the connection pool if an error happened in the caller. They might contain unread messages that could corrupt the data for the next request.

**Problem location**
Rawclient

Invalidate the connection when error happens.

Tristan1900:
**Change log description**  
Fix the `payLoadLengthTooLong` issue.
Prevent connections that has stale data from being dropped back into the pool.

**Purpose of the change**  
Fix #264

**What the code does**  
Marks connection as invalid when created. `ClientConnection` later validates the connection when it consumes all the data in the connection. This way we can be sure that every connection that's dropped back into the pool is clean and it doesn't have corrupted data.

**How to verify it**  
Tests should pass, also need to verify in the gstreamer experiment.


wire_protocol/src/client_connection.rs
```diff
@@ -117,6 +117,7 @@ impl ClientConnection for ClientConnectionImpl<'_> {
 }
 
 pub async fn read_wirecommand(connection: &mut dyn Connection) -> Result<Replies, ClientConnectionError> {
+    connection.set_validity(false);
```
tkaitchuck, line 120: Right now we have two concepts with very similar names. I don't know if they can be combined. 
See line 108 above. It is calling `connection.invalidate()` which is distinct from `connection.set_validity(false)` if these can't be made into one thing, perhaps one or both should be renamed.

Tristan1900: thanks Tom, good point!
I have removed the old logic as the new logic is doing the same job.

wire_protocol/src/connection.rs
```diff
@@ -227,7 +237,13 @@ impl Connection for TlsConnection {
     }
 
     fn is_valid(&self) -> bool {
-        self.stream.as_ref().expect("get connection").is_valid()
+        self.is_valid
+            && self.stream.as_ref().is_some()
```
shrids, line 241: when can `self.stream.as_ref()` be None?

Tristan1900: the connection can be splited into writeHalf and readHalf. After the split the original stream is None.

wire_protocol/src/connection.rs
```diff
@@ -227,7 +237,13 @@ impl Connection for TlsConnection {
...
+            && self.stream.as_ref().expect("get connection").is_valid()
+    }
+
+    fn set_validity(&mut self, is_valid: bool) {
```
tkaitchuck, line 245: A comment might be in order, because this leads to the somewhat surprising case that calling `set_validity(true)` followed by `is_valid` does not always return `true`.

tkaitchuck (approved):

shrids (approved):

PR #268, opened 2021-05-31
Issue 247: Accept scheme in controller URI

### LINKED ISSUE #247: Accept controller URI in tcp://ip:port format
**Problem description**
`ClientConfig.controller_uri` currently requires the value to be in a format such as `1.2.3.4:9090` or `controller.example.com:443`. This is not a true URI since it is missing the schema (see https://en.wikipedia.org/wiki/Uniform_Resource_Identifier), is inconsistent with the Java API, and requires a separate parameter to enable TLS. 

**Problem location**
ClientConfig

Accept controller_uri in formats `tcp://1.2.3.4:9090` and `tls://controller.example.com:443`. This change should be made in a backwards-compatible way.

bkc:
**Change log description**  
* allow scheme to be included in PravegaNodeUri
* changes ClientConfigBuilder to automatically set `is_tls_enabled = true` when uri scheme is 'tls'
* allows tls scheme to be used with StreamManager Python client when tls_enabled default value (false) is used

**Purpose of the change**  

Fixes #247 

**What the code does**  
changes to shared/lib
* accepts schemes in PravegaNodeUri
* added PravegaNodeUriParseError
* added PravegaNodeUri::is_well_formed static method
* added PravegaNodeUri::uri_parts_from_string static method
* changed existing methods to use uri_parts_from_string

changes to config/lib
* added default builder for `ClientConfigBuilder::is_tls_enabled`
* altered `ClientConfigBuilder::extract_trustcerts` to also call `default_is_tls_enabled` (builder order of operations calls extract_trustcerts before default_is_tls_enabled)
* added `ClientConfigBuilder::validate` to check for conflicts between uri scheme and `.is_tls_enabled()`
* removed duplicate `builder(setter(into))`  from controller_uri because its already set at the struct level

changes to python_binding/stream_manager
* changed `StreamManager` constructor, if `tls_enabled = false` then rely on the scheme from `controller_uri` to set `is_tls_enabled`
* updated sample usage

**How to verify it**  

* rust tests have been added 
* was not able to test python changes due to issues with maturin

**Comments**

The only uri scheme this code specifically tests for is `tls`. It should probably only allow `tcp, tls` schemes (but see my outstanding question on #247)

Some existing PravegaNodeUri methods panic if the uri is malformed. I added `is_well_formed` static method so that library consumers could test a uri before building a clientconfig.  This approach seems a bit dodgy. Perhaps changing PravegaNodeUri to a struct not derived from String could improve this, but I think that would require more changes to the library.

If we want true backwards compatibility with the python client, we may wish to change StreamManager `tls_enabled` arg to `PyAny` type, then set the default python arg to either `None` or a sentinel object that can make it clear whether or not the python caller provided a value for `tls_enabled`

This is my first rust code. Some of the changes don't look quite right. I considered using Regex but its currently only used in the auth crate, so didn't seem worth pulling into this one.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -47,9 +47,10 @@ const AUTH_PROPS_PREFIX_ENV: &str = "pravega_client_auth_";
 
 const TLS_CERT_PATH_ENV: &str = "pravega_client_tls_cert_path";
 const DEFAULT_TLS_CERT_PATH: &str = "./certs";
+const TLS_SCHEME: &str = "tls";
 
 #[derive(Builder, Debug, Getters, CopyGetters, Clone)]
-#[builder(setter(into))]
+#[builder(setter(into), build_fn(validate = "Self::validate"))]
 pub struct ClientConfig {
     #[get_copy = "pub"]
     #[builder(default = "u32::max_value()")]
@@ -68,7 +69,6 @@ pub struct ClientConfig {
     pub retry_policy: RetryWithBackoff,
 
     #[get]
-    #[builder(setter(into))]
     pub controller_uri: PravegaNodeUri,
 
     #[get_copy = "pub"]
@@ -79,12 +79,12 @@ pub struct ClientConfig {
     #[builder(default = "false")]
     pub mock: bool,
 
+    #[get_copy = "pub"]
+    #[builder(default = "self.default_is_tls_enabled()")]
+    pub is_tls_enabled: bool,
+
     #[builder(default = "self.extract_trustcerts()")]
     pub trustcerts: Vec<String>,
-
-    #[get_copy = "pub"]
-    #[builder(default = "false")]
-    pub is_tls_enabled: bool,
 
     #[builder(default = "self.extract_credentials()")]
     pub credentials: Credentials,
@@ -108,8 +108,10 @@ impl ClientConfigBuilder {
             if !enable {
                 return vec![];
             }
-        } else {
+        } else if !self.default_is_tls_enabled() {
             return vec![];
+        } else {
+            // fall through to parsing certs path
         }
         let ret_val = env::vars()
             .filter(|(k, _v)| k.starts_with(TLS_CERT_PATH_ENV))
@@ -157,6 +159,47 @@ impl ClientConfigBuilder {
 
     fn default_timeout(&self) -> Duration {
         Duration::from_secs(30)
+    }
+
+    fn default_is_tls_enabled(&self) -> bool {
+        if let Some(controller_uri) = &self.controller_uri {
+            return match controller_uri.scheme() {
+                Ok(scheme) => scheme == TLS_SCHEME,
+                Err(_) => false,
+            };
+        }
+        false
+    }
+    /// validate the builder before returning it
+    ///
+    /// if both is_tls_enabled and controller_uri have been set
+    /// verify that the uri scheme (if present) matches the is_tls_enabled
+    /// value.
+    fn validate(&self) -> Result<(), String> {
+        if self.is_tls_enabled.is_none() || self.controller_uri.is_none() {
+            // at least one option has not been specified, cannot have a conflict
+            return Ok(());
+        }
+        let is_tls_enabled = self.is_tls_enabled.unwrap();
+        let scheme_result = self.controller_uri.as_ref().unwrap().scheme();
+        if scheme_result.is_err() {
+            // unparsable URI, cannot have a conflict
+            // though maybe we should return the Err value
+            return Ok(());
+        }
+        let scheme = scheme_result.unwrap();
+        if scheme.is_empty() {
+            // no scheme specified, cannot have a conflict
+            return Ok(());
+        };
+
+        match (is_tls_enabled, scheme == TLS_SCHEME) {
+            (true, true) | (false, false) => Ok(()),
+            _ => Err(format!(
+                "is_tls_enabled option {} does not match uri scheme {}",
+                is_tls_enabled, scheme
+            )),
+        }
     }
 }
 
@@ -249,7 +292,7 @@ mod tests {
     #[test]
     fn test_extract_tls_cert_path() {
         // test default
-        fs::create_dir(DEFAULT_TLS_CERT_PATH).expect("create default cert path");
+        fs::create_dir_all(DEFAULT_TLS_CERT_PATH).expect("create default cert path");
         fs::File::create(format!("{}/foo.crt", DEFAULT_TLS_CERT_PATH)).expect("create crt");
         let config = ClientConfigBuilder::default()
             .controller_uri("127.0.0.2:9091".to_string())
@@ -257,10 +300,39 @@ mod tests {
             .build()
             .unwrap();
         assert_eq!(config.trustcerts.len(), 1);
-        fs::remove_dir_all(DEFAULT_TLS_CERT_PATH).expect("remove dir");
 
+        // test w/ tls uri prefix
+        let config = ClientConfigBuilder::default()
+            .controller_uri(format!("{}://127.0.0.2:9091", TLS_SCHEME))
+            .build()
+            .unwrap();
+        assert_eq!(config.trustcerts.len(), 1);
+
+        // test w/o tls uri prefix
+        let config = ClientConfigBuilder::default()
+            .controller_uri("tcp://127.0.0.2:9091".to_string())
+            .build()
+            .unwrap();
+        assert_eq!(config.trustcerts.len(), 0);
+
+        // test conflicting tls setting vs scheme
+        let conflicted_config1 = ClientConfigBuilder::default()
+            .controller_uri(format!("{}://127.0.0.2:9091", TLS_SCHEME))
+            .is_tls_enabled(false)
+            .build();
+        assert!(conflicted_config1.is_err());
+
+        // test alternate conflicting tlst setting vs scheme
+        let conflicted_config2 = ClientConfigBuilder::default()
+            .controller_uri("tcp://127.0.0.2:9091".to_string())
+            .is_tls_enabled(true)
+            .build();
+
+        assert!(conflicted_config2.is_err());
+
+        fs::remove_dir_all(DEFAULT_TLS_CERT_PATH).expect("remove dir");
         // test with env var set
-        fs::create_dir("./bar").expect("create default cert path");
+        fs::create_dir_all("./bar").expect("create default cert path");
         fs::File::create(format!("./bar/foo.crt")).expect("create crt");
         env::set_var("pravega_client_tls_cert_path", "./bar");
         let config = ClientConfigBuilder::default()
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -47,7 +47,7 @@ const AUTH_PROPS_PREFIX_ENV: &str = "pravega_client_auth_";
 
 const TLS_CERT_PATH_ENV: &str = "pravega_client_tls_cert_path";
 const DEFAULT_TLS_CERT_PATH: &str = "./certs";
-const TLS_SCHEME: &str = "tls";
+const TLS_SCHEMES: [&str; 4] = ["tls", "ssl", "tcps", "pravegas"];
 
 #[derive(Builder, Debug, Getters, CopyGetters, Clone)]
 #[builder(setter(into), build_fn(validate = "Self::validate"))]
@@ -164,7 +164,7 @@ impl ClientConfigBuilder {
     fn default_is_tls_enabled(&self) -> bool {
         if let Some(controller_uri) = &self.controller_uri {
             return match controller_uri.scheme() {
-                Ok(scheme) => scheme == TLS_SCHEME,
+                Ok(scheme) => TLS_SCHEMES.contains(&&*scheme),
                 Err(_) => false,
             };
         }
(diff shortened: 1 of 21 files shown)


shrids (approved):

config/src/lib.rs
```diff
@@ -68,7 +69,6 @@ pub struct ClientConfig {
     pub retry_policy: RetryWithBackoff,
 
     #[get]
-    #[builder(setter(into))]
```
tkaitchuck, line 71: Is this being removed because it is redundant?

config/src/lib.rs
```diff
@@ -108,8 +108,10 @@ impl ClientConfigBuilder {
             if !enable {
                 return vec![];
             }
-        } else {
+        } else if !self.default_is_tls_enabled() {
             return vec![];
+        } else {
```
tkaitchuck, line 113: Else is not needed

config/src/lib.rs
```diff
@@ -158,6 +160,47 @@ impl ClientConfigBuilder {
...
+    ///
+    /// if both is_tls_enabled and controller_uri have been set
+    /// verify that the uri scheme (if present) matches the is_tls_enabled
+    /// value.
+    fn validate(&self) -> Result<(), String> {
+        if self.is_tls_enabled.is_none() || self.controller_uri.is_none() {
+            // at least one option has not been specified, cannot have a conflict
+            return Ok(());
+        }
+        let is_tls_enabled = self.is_tls_enabled.unwrap();
```
tkaitchuck, line 183: Per the documentation:
> Default values are applied after validation, and will therefore not be validated!

So `is_tls_enabled` may be none, so unwrapping here is not a good idea. This should really use a `.unwrap_or(false)`

config/src/lib.rs
```diff
@@ -158,6 +160,47 @@ impl ClientConfigBuilder {
...
+        let scheme_result = self.controller_uri.as_ref().unwrap().scheme();
```
tkaitchuck, line 184: Same here and below.

python_binding/src/stream_manager.rs
```diff
@@ -30,6 +30,9 @@ cfg_if! {
 /// manager=pravega_client.StreamManager("127.0.0.1:9090")
 /// // this manager can be used to create scopes, streams, writers and readers against Pravega.
 /// manager.create_scope("scope")
+///
+/// // optionally enable tls support using tls:// scheme
```
tkaitchuck, line 34: Let's change the non-tls example to use the 'tcp' prefix so that it's consistent.

shared/src/lib.rs
```diff
@@ -50,6 +51,21 @@ extern crate shrinkwraprs;
...
 
+type PravegaNodeUriParseResult<T> = std::result::Result<T, PravegaNodeUriParseError>;
+
+#[derive(Debug, Snafu)]
+pub enum PravegaNodeUriParseError {
+    #[snafu(display("Could not parse uri due to {}", error_msg))]
+    ParseError { error_msg: String },
+}
+
+#[derive(Debug, PartialEq, Default)]
```
tkaitchuck, line 62: Do we actually want 'default'? I don't see how it does much good to construct one that's all `none`.

shared/src/lib.rs
```diff
@@ -50,6 +51,21 @@ extern crate shrinkwraprs;
...
+struct PravegaNodeUriParts {
+    scheme: Option<String>,
+    authority: Option<String>,
```
tkaitchuck, line 65: Is this actually optional?

shared/src/lib.rs
```diff
@@ -75,22 +91,87 @@ impl From<SocketAddr> for PravegaNodeUri {
...
-        let mut addrs_vec: Vec<_> = self
-            .0
-            .to_socket_addrs()
-            .expect("Unable to resolve domain")
-            .collect();
-        addrs_vec.pop().expect("get the first SocketAddr")
+        match PravegaNodeUri::uri_parts_from_string(self.to_string()) {
+            Ok(uri_parts) => {
+                let mut addrs_vec: Vec<_> =
+                    format!("{}:{}", uri_parts.authority.unwrap(), uri_parts.port.unwrap())
```
tkaitchuck, line 97: Two points:
1. Per: https://uri.thephpleague.com/uri/4.0/components/overview/ athority includes the port. Should this be `host`?
2. Avoid `unwrap` here. We don't want to panic. Ideally the object would be structured such that such cases are impossible at runtime. If we absolutely can't avoid it, then use `.expect`

shared/src/lib.rs
```diff
@@ -75,22 +91,87 @@ impl From<SocketAddr> for PravegaNodeUri {
...
+            Err(_) => false,
+        }
+    }
+
+    fn uri_parts_from_string(uri: String) -> PravegaNodeUriParseResult<PravegaNodeUriParts> {
+        let mut uri_parts: PravegaNodeUriParts = PravegaNodeUriParts::default();
+        let mut parts: Vec<_> = uri.split("://").collect();
+        if parts.len() > 2 || parts.is_empty() {
+            return Err(PravegaNodeUriParseError::ParseError {
+                error_msg: "malformed uri".into(),
```
tkaitchuck, line 146: Here and below include the bad URI in the error message.

tkaitchuck (changes requested):

config/src/lib.rs
bkc: Yes, it was removed because it duplicates the setting on the builder struct

config/src/lib.rs
bkc: I agree with you, but clippy "else_if_without_else" disagrees with both of us.

I applied the workaround recommended at https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/else_if_without_else.rs

config/src/lib.rs
bkc: `self.controller_uri.as_ref().unwrap()` is also guarded at line 179.  `scheme()` returns a String. I will also change this to re-check as requested.

config/src/lib.rs
bkc: unwrap cannot fail here due to the guard at line 179 `self.is_tls_enabled.is_none()`

I suppose a future change could remove that guard. I will apply the change you requested.

shared/src/lib.rs
bkc: These members are Optional because their corresponding components may not be present when `uri_parts_from_string` decodes the string.

shared/src/lib.rs
bkc: I was being a bit lazy here. Authority also contains userinfo, which I did not want to parse

I could change the name `authority` to `userinfo_host`, to be clearer?

bkc: regarding the unwrap and panic. The existing `port()` method doesn't check for possible panic due to a non-numeric port number.  I did not want to change the existing semantics on how `port()` works, or the existing expectation that `port` had to be valid to avoid a panic.

Interestingly it appears that `port()` is only used in tests. If we can remove all uses of `port()`, we could  split the URI into just scheme and authority and leave the panic to `SocketAddr`. which would most likely happen at the time of  connection use instead of during config build.

shared/src/lib.rs
bkc: are you saying the error message will include the uri, or that it should include the uri?

I don't think the uri is currently being included in the error message. 

Are you sure you want the uri to be included in the error message. What if it includes userinfo (future functionality?)

shared/src/lib.rs
bkc: PravegaNodeUri has public fn `domain_name()` which is used elsewhere in controller-client and connection_factory.  PravegaNodeUri appears to be used to parse a host name returned from the controller and to associate tls certificates with a host name.

The `domain_name` function returns the value I lazily named `authority`.  The existing `domain_name` function did not parse the uri to extract only the host portion of a URI. That is, it didn't exclude userinfo, which might be provided by an end-user when building a connection (future feature) but probably won't be returned by a controller.

Would it be reasonable to ignore the userinfo issue for now?  

Should field 'authority' be renamed 'domain_name' to match the existing function name, or should function `domain_name` be renamed to `host()` to match the URI specification terminology, or something else?

Tristan1900 (approved):

shared/src/lib.rs
tkaitchuck: I think it should be included to aid debugging. We don't have userinfo now, and this would only be returning the string that was passed in back to the code that provided it, so I don't see an issue.

bkc: k, I will get back to the PR  this weekend, will add this.

shared/src/lib.rs
tkaitchuck: As we discussed, I think we can split this into two different objects, one internal to refer to segment store instances and one external to be included in config to refer to controller instances. 
For the external one, we should parse early and return errors or panic if it is malformed. For the internal one we can do as before and have a simple string wrapper without a scheme.

shared/src/lib.rs
tkaitchuck: I think we can ignore userinfo. We don't use it, and probably never will.

shared/src/lib.rs
bkc: for external URI, it appears all the existing code doesn't seem to expect an error when the URI itself is constructed.  Instead, it appears that only `build()`  could return an Err. When you say "parse early", do you mean literally when the URI is constructed, or when `build()` is called on the ClientConfigBuilder?


e.g. this example

```rust
    let config = ClientConfigBuilder::default()
        .controller_uri("localhost:9090")
        .build()
        .unwrap();
```

Would it be reasonable to wait until `build()` is called before parsing the value passed to `controller_uri`?
I don't believe `.controller_uri()` can return an Err, so it could only panic if you wanted to parse the URI at the time it is instantiated.

tkaitchuck: Probably best to do it in the validation of the config. That is where an error is most expected.

shared/src/lib.rs
bkc: I renamed `authority` to `domain_name` to match other parts of the code base. Its not precise, but it is "same name for the same thing"

config/src/lib.rs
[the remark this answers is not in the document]
bkc: I changed the `validate` method a lot, this suggestion no longer fits.

config/src/lib.rs
```diff
@@ -158,6 +160,47 @@ impl ClientConfigBuilder {
...
+            return Ok(());
+        };
+
+        match (is_tls_enabled, scheme == TLS_SCHEME) {
+            (true, true) | (false, false) => Ok(()),
+            _ => Err(format!(
+                "is_tls_enabled option {} does not match uri scheme {}",
+                is_tls_enabled, scheme
+            )),
+        }
```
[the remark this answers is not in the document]
bkc: heh.. I applied your simpler construct

config/src/lib.rs
```diff
@@ -158,6 +160,47 @@ impl ClientConfigBuilder {
...
+            return Ok(());
+        }
+        let is_tls_enabled = self.is_tls_enabled.unwrap();
+        let scheme_result = self.controller_uri.as_ref().unwrap().scheme();
+        if scheme_result.is_err() {
+            // unparsable URI, cannot have a conflict
+            // though maybe we should return the Err value
+            return Ok(());
+        }
+        let scheme = scheme_result.unwrap();
```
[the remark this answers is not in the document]
bkc: I changed `validate()` to  re-use existing method `default_is_tls_enabled()` though I expect it could use further simplification

shared/src/lib.rs
bkc: I have added uri to the error messages.

shared/src/lib.rs
bkc: validation is done in `build()` step. 

I could not find where `segment store`  endpoints are returned from the controller, so I could not create a new/simpler, endpoint/uri type for that use case.  `ConnectionFactory` seems like where an endpoint would be used, but I could not figure out which call(s) to `establish_connection` would need to change to use the new type.

bkc:
can someone please approve a workflow/check run so I can be sure there aren't any test/clippy issues with this PR?

Thanks

shared/src/lib.rs
```diff
@@ -75,22 +91,101 @@ impl From<SocketAddr> for PravegaNodeUri {
...
+        let first_endpoint = match uri.split(',').next() {
+            Some(endpoint) => endpoint,
+            _ => {
+                return Err(PravegaNodeUriParseError::ParseError {
+                    error_msg: format!("malformed uri {}", uri),
+                })
+            }
+        };
+
+        let mut scheme_authority: Vec<_> = first_endpoint.split("://").collect();
```
tkaitchuck, line 157: While by no means a requirement, I'd be tempted to use [regex](https://docs.rs/regex/1.5.4/regex/):
```regex
(?x)
((?:P<scheme>[[:alnum:]-]*)://)?
(?:P<host>[[:alnum:]-]*)):
(?:P<port>[[:digit:]]*)
```

tkaitchuck (approved):

shared/src/lib.rs
bkc: Hi Tom,

In my original PR comment I said:

> I considered using Regex but its currently only used in the auth crate, so didn't seem worth pulling into this one.

At that time I wasn't sure what the ramifications were of using regex in this crate.

I will go ahead and change it.

bkc:
@shrids Hi, can you or someone else please approve a workflow run so I can be certain this PR is complete. Thanks

shrids (approved):

shrids:
@bkc the changes look good 👍 . Could you fix the Clippy failure here ?

bkc:
> @bkc the changes look good  . Could you fix the Clippy failure here ?

Sure.

It's too bad the workflow quits at the first clippy issue. There are several more than this one in other modules. I'll resolve them all.

bkc:
all clippy issues should be resolved.

bkc:
workspace level clippy issues should be resolved.

PR #275, opened 2021-06-08
Issue 274: Remove openssl

### LINKED ISSUE #274: Remove openssl in dependency
**Problem description**
Reqwest crate is using openssl by default. We should use rustls since we are using rustls for security already, using openssl will add complexity to our dependency and increase the library size.

**Problem location**
cargo.toml file

Tristan1900:
**Change log description**  
Reqwest is using openssl by default, we can change it to use rustls.

**Purpose of the change**  
Fix #274

**What the code does**  
Use rustls instead of openssl.

**How to verify it**  
current test should pass

### DIFF
--- a/config/Cargo.toml
+++ b/config/Cargo.toml
@@ -20,7 +20,7 @@ derive_builder = "0.9"
 getset = "0.0.9"
 serde_json = "1.0"
 serde = "1.0"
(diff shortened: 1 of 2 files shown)


shrids (approved):
LGTM.

PR #276, opened 2021-06-09
Issue 222: Async byte stream api

### LINKED ISSUE #222: Async API for ByteStream
**Problem description**
Provide async API for ByteStream so application can use their own executor.

**Problem location**

Tristan1900:
**Change log description**  
Add async API for reader and writer. Instead of implementing `futures::asyncWrite` and `futures::asyncReader` trait, we just provide a `write_async` and `read_async` API.

**Purpose of the change**  
Fix #222

**What the code does**  
Add async verions of `write`, `seek_to_tail`, `read`, `seek`.

**How to verify it**  
Tests should pass


src/byte/writer.rs
```diff
@@ -152,11 +158,28 @@ impl ByteWriter {
...
+    /// Write data asynchronously.
+    ///
+    /// # Examples
+    /// ```ignore
+    /// let mut byte_writer = client_factory.create_byte_writer_async(segment).await;
+    /// let payload = vec![0; 8];
+    /// let (size, handle) = byte_writer.write(&payload).await;
+    /// assert!(handle.await.is_ok());
+    /// ```
+    pub async fn write_async(&mut self, buf: &[u8]) -> (usize, oneshot::Receiver<Result<(), Error>>) {
```
tkaitchuck, line 170: The default path is already buffered and we don't have acks. So what if, rather than modeling it on the event path which has oneshots, and the caller does `.await.await`, we just have `write_async` and `flush_async` which are both identical to `write` and `flush` except they don't block. This avoid the need for the oneshots.

integration_test/src/byte_reader_writer_tests.rs
```diff
@@ -98,17 +95,31 @@ pub fn test_byte_stream(config: PravegaStandaloneServiceConfig) {
...
+    ));
+    handle.block_on(test_async_write_and_read(&client_factory, segment));
 }
 
 fn test_simple_write_and_read(writer: &mut ByteWriter, reader: &mut ByteReader) {
     info!("test byte stream write and read");
     let payload1 = vec![1; 4];
     let payload2 = vec![2; 4];
-
+    // writer.seek_to_tail();
```
shrids, line 118: should we uncomment this?

tkaitchuck (approved):

PR #277, opened 2021-06-17
Issue 234: Implement index stream 

### LINKED ISSUE #234: Create an api for "Index streams".
Indexes are a common need for video.

Below is a proposal for how they could be implemented:

Proposal
========

An index stream is a stream where entries are added which have a specific structure.
This structure consists of a fixed number of monotonically increasing values, and some user defined data.
So for example a user could define three values: "time", "sequence number", "location" and define each of these with a long. 
(All of the defined keys have a value for every append)
Then every time an entry is written a check would be made to make sure each of these is increasing.
If these conditions are met the new entry (the values and any associated data) are appended to the stream.
This write is atomic.

APIs
========
* Get the first value. - For any of the user's defined keys return the oldest untruncated value in the stream.
* Get latest value - For any of the user's defined keys return the most recent one in the stream. 
* Add new key - All future entries are expected to provide the given key or fail.
* Append data - (described above)
* Append data conditional on any of the values being the value provided.
* Read data starting at a Revision (optionally up to another Revision)
* Find the first entry (and revision) with a value >= x (or <= x)
* Start reading from the first entry with a value >= x (or <= x)
* Truncate to a revision

Implementation
========

One way to implement this efficiently is to have a fixed size record. For example 4kb.
Then we can enforce a limit the size by for example no more having than 100 entries, and no more than 1kb of custom data.
Then all writes are written as conditional appends based on offset. 

This way it is always possible to align a read to a 4kb boundary and know that it is aligned to a record. So binary (or newtonian) search can be performed by aligning reads to a 4kb boundary and reading 4kb.

To allow the keys to change over time, it may be best to simply insert the keys that are used at the start of the record. Then if the keys change a new frame needs to be started (padding out as needed).

Tristan1900:
**Change log description**  
Add support to the index stream.

**Purpose of the change**  
Fix #234 

**What the code does**  
***Naming***
* Record: Fixed size(4KB) event written to the stream. A record contains: type code(4 bytes), entries length(4 bytes), data length(4 bytes) entries, entries and data.
* Field: A key value pair(&'static str, u64) . Key is a string that user defines, value is a u64 that is monotonically increasing.
* Fields: A set of field. User specifies a label struct that has a number of fields, the name of the field is the entry key and the value is entry value. This impl provides a procedure macro called label that can parse the struct to an array of entries. 

***Index writer***:
* append(data, label): append the given label and data to the stream. It is guaranteed that the new label has all the exsiting entries and their values are not less than the current values.
* append_conditionally(data, label, condition_label): same as above, plus a given condition label that has to be exactly the same as the latest label in the stream
* flush(): flush appends
* truncate(offset): truncate to a offset.

***Index reader***:
* search_offset(entry: (key, value)): search for the offset of the first record that has a matching key and value that is equal to or larger than the target value 
* read(start_offset, end_offset): read a slice of a stream. When end_offset is u64::MAX, it means tail read.
* first_record(): first readable record
* last_record(): latest record
* head_offset():
* last_offset():

**How to verify it**  
Tests should pass.


src/index/writer.rs
```diff
@@ -0,0 +1,235 @@
...
+        IndexWriter {
+            byte_writer,
+            entries: None,
+            label: None,
+            _label_type: PhantomData,
+        }
+    }
+
+    /// Append data with a given label
+    pub async fn append(&mut self, data: Vec<u8>, label: T) -> Result<(), IndexWriterError> {
```
tkaitchuck, line 126: Let's reverse this parameter order and have labels first. (Same elsewhere)

src/index/writer.rs
```diff
@@ -0,0 +1,235 @@
...
+        let index_reader = factory.create_index_reader(segment.clone()).await;
+        let tail_offset = index_reader.tail_offset().await.expect("get tail offset");
+        let head_offset = index_reader
+            .head_offset()
+            .await
+            .expect("get readable head offset");
+        if head_offset != tail_offset {
+            let prev_record_offset = tail_offset - RECORD_SIZE;
+            let record = index_reader
+                .read_record_from_random_offset(prev_record_offset)
```
tkaitchuck, line 106: Won't `last_record` do this in a simpler way?

src/index/writer.rs
```diff
@@ -0,0 +1,235 @@
...
+                .expect("read last record");
+
+            return IndexWriter {
+                byte_writer,
+                entries: Some(record.entries),
+                label: None,
+                _label_type: PhantomData,
+            };
+        }
+        IndexWriter {
```
tkaitchuck, line 117: This is the same inside and outside the `if`. Maybe just has a local variable for `entries`?

macros/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+
+use proc_macro2::TokenStream;
+use quote::{quote, quote_spanned};
+use syn::spanned::Spanned;
+use syn::{parse_macro_input, parse_quote, Data, DeriveInput, Fields, GenericParam, Generics};
+
+#[proc_macro_derive(Label)]
+pub fn derive_label(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
shrids, line 17: Could you please add documentation for this ?

src/index/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    fn read_from(input: &[u8]) -> Result<Self, BincodeError> {
+        let decoded: Record = CONFIG.deserialize(&input[8..])?;
+        Ok(decoded)
+    }
+}
+
+pub(crate) fn hash_key_to_u128(key: &'static str) -> u128 {
+    let mut shake = Shake::v128();
```
shrids, line 110: Can you share the doc link for this library here?

Tristan1900: Sure, https://docs.rs/tiny-keccak/2.0.2/tiny_keccak/struct.Shake.html as Tom suggests https://github.com/pravega/pravega-client-rust/issues/234#issuecomment-854095681

shrids (approved):

src/index/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+// Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+
+//! The Index API provides a way to efficiently search data in the stream.
```
tkaitchuck, line 11:
```suggestion
//! The Index API provides a way to build a monotonic index with a stream.
```

src/index/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+//! struct MyFields {
+//!     time: u64, // A field
+//!     id: u64,
+//! }
+//!
+//! ```
+//!
+//! To ensure the searching efficiency in an index stream, we impose some constraints to the `Fields`:
+//! * The value of the `Field` must be monotonically increasing.
+//! * Deletion of the `Field` is not permitted.
```
tkaitchuck, line 30: This should be put into a section about versioning and upgrades. Here it sounds like the structure itself is dynamic.

src/lib.rs
```diff
@@ -26,6 +26,7 @@
 )]
 #![allow(clippy::multiple_crate_versions, clippy::needless_doctest_main)]
 #![allow(bare_trait_objects)]
+#![recursion_limit = "1024"]
```
tkaitchuck, line 29: What's going on here?

Tristan1900: the default value is 256 and `futures:stream` macro in the `read` method expands more than that so cargo clippy won't pass.

src/index/reader.rs
```diff
@@ -0,0 +1,283 @@
...
+            msg: format!("error when fetching head offset: {:?}", e),
+        })? as i64;
+        let tail = self.tail_offset().await.map_err(|e| IndexReaderError::Internal {
+            msg: format!("error when fetching tail offset: {:?}", e),
+        })? as i64;
+        let mut start = 0;
+        let num_of_record = (tail - head) as i64 / RECORD_SIZE_SIGNED;
+        let mut end = num_of_record - 1;
+
+        while start <= end {
```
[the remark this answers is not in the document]
Tristan1900: Yeah I'm not sure how to add unit test for this one. Some other components are involved in this method and I couldn't fin an easy way to mock them. I do add integration test to test the search_offset logic though.

src/index/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+        config.big_endian();
+        config.limit(RECORD_SIZE);
+        config.array_length(LengthOption::U32);
+        config.string_length(LengthOption::U16);
+        config
+    };
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct IndexRecord {
```
tkaitchuck, line 87: Does this need to be pub?

tkaitchuck (approved):

PR #279, opened 2021-06-18
Issue 278: Fix segment number

### LINKED ISSUE #278: Segment number does not start from 0
**Problem description**
If a stream recreated right after deletion, the segment number may start from a non-zero number due to the lazy deletion of segment in tier2. So if user provides a segment 0 to the Byte Writer, what it means is actually segment x.

**Problem location**
Byte Writer

Tristan1900:
**Change log description**  
Byte stream API now takes a fixed size stream instead of a paticular segment as the input. This stream is a special stream that could only contains one segment.

**Purpose of the change**  
FIx #278

**What the code does**  
Use stream instead of segment as byte stream API input. Ensure that the stream contains only one segment.

**How to verify it**  
Tests should pass.


tkaitchuck (changes requested):
I don't think we want to adjust the number given to us by users. We should instead just take the stream name and work it out from the controller.

shrids (approved):
LGTM apart from the two errors in the tests.

integration_test/src/lib.rs
```diff
@@ -94,12 +94,12 @@ mod test {
             .unwrap();
 
         // metric::metric_init(PROMETHEUS_SCRAPE_PORT.parse::<SocketAddr>().unwrap());
-        info!("Running integration test");
-        let config = PravegaStandaloneServiceConfig::new(false, true, true);
-        run_tests(config);
-
-        let config = PravegaStandaloneServiceConfig::new(true, false, false);
-        run_tests(config);
+        // info!("Running integration test");
```
tkaitchuck, line 97: What's going on here?

tkaitchuck (commented):
This looks good. But it's changing a public API so you'll need to bump the version number in the toml.

integration_test/src/lib.rs
Tristan1900: fixed!

tkaitchuck (approved):

PR #281, opened 2021-06-30
Issue 280: Enable Stream tags for Python bindings.

### LINKED ISSUE #280: Enable Stream tags for Python bindings.
**Problem description**
Enable Stream tags for Python bindings.

**Problem location**
Python bindings

Fix it.

shrids:
**Change log description**  
- Enable Stream tags for Python bindings.

**Purpose of the change**  
Fixes #280 

**What the code does**  
- Enable passing of Stream scaling policy and Retention policy for python bindings.
- Enable Stream Tag APIs on python bindings.

```python

# create a stream with stream scaling is enabled with data rate as 10kbps, scaling factor as 2 and initial segments as 1
scalingpolicy = StreamScalingPolicy.auto_scaling_policy_by_data_rate(10, 2, 1)
stream_manager.create_stream_with_policy(scope_name=scope, stream_name="testStream1", scaling_policy=scalingpolicy)
# By default retention policy is disabled.

# update stream with retentionpolicy and tags.
##retention policy of 10GB
retentionpolicy = StreamRetentionPolicy.by_size(10*1024*1024 * 1024)
stream_manager.update_stream_with_policy(scope, "testStream1", scalingpolicy, retentionpolicy, tags=['t1', 't2'])

# fetch stream tags for the given Stream
tags = stream_manager.get_stream_tags(scope, "testStream1")
```

**How to verify it**  


shared/src/lib.rs
```diff
@@ -458,8 +458,8 @@ impl Default for Scaling {
         Scaling {
             scale_type: ScaleType::FixedNumSegments,
             min_num_segments: 1,
-            scale_factor: 1,
-            target_rate: 1000,
+            scale_factor: 0,
```
tkaitchuck, line 461: Is 0 a valid number here?

tkaitchuck (approved):

python_binding/tests/pravega_client_test.py
```diff
@@ -13,9 +13,46 @@
 import string
 import pravega_client
 from pravega_client import TxnFailedException
+from pravega_client import StreamScalingPolicy
+from pravega_client import StreamRetentionPolicy
 
 class PravegaTest(unittest.TestCase):
 
+    def test_tags(self):
```
Tristan1900, line 21: Probably we need to add a test case when there are no tags, then the result should be empty.

Tristan1900 (approved):

shared/src/lib.rs
shrids: It is a no-op since the scaling type is `ScaleType::FixedNumSegments`.
I have changed it to 1.

python_binding/tests/pravega_client_test.py
shrids: done, improve the test.

PR #283, opened 2021-06-30
Issue 282: Add config to skip cert verification

### LINKED ISSUE #282: Add configuration for allowing self-signed certificate.
**Problem description**
Currently, when auth and tls are enabled, authentication will use `reqwest` to fetch access token from keycloak. `reqwest` itself will verify if the tls certs are valid, so any self-signed certs cannot pass this verfication.

**Problem location**
Authentication

`reqwest` provides a configuration that can disable the verfication. We should also support such configuration.

Tristan1900:
**Change log description**  
Add a config to skip cert verification in `reqwest`.

**Purpose of the change**  
Fix #282 

**What the code does**  
Add a config to skip the cert verification in `reqwest`.
`StreamManager` in Python binding now have a new paramter in new method: 
```
    fn new(controller_uri: &str, auth_enabled: bool, tls_enabled: bool,disable_cert_verification: bool)
```
**How to verify it**


shrids (approved):
LGTM on code inspection.

PR #288, opened 2021-07-07
Issue 291: Config also accepts cert file path

Tristan1900:
**Change log description** 
Application can also pass cert file path to the config. Right now it only accept cert directory path.

**What the code does**  
Check the path type first. If it is a directory, then read all the files in it. If it is a file, then just read the file.

**How to verify it**  
Test should pass.


config/src/credentials.rs
```diff
@@ -57,7 +57,7 @@ impl Credentials {
         }
     }
 
-    pub fn keycloak(path: &str) -> Self {
+    pub fn keycloak(path: &str, skip_cert_verification: bool) -> Self {
```
tkaitchuck, line 60: Nit: Here and elsewhere, can we call it "Disable" cert verification?

tkaitchuck (approved):

shrids (approved):
LGTM..

claudiofahey (approved):

PR #290, opened 2021-07-08
Issue 292: Fix transaction id format error

Tristan1900:
**Change log description**  
The hexadecimal based transaction id was using upper case letter whereas lower case letter should be used.

**What the code does**  
use lower case letter

**How to verify it**  
test should pass

### DIFF
--- a/integration_test/src/transactional_event_writer_tests.rs
+++ b/integration_test/src/transactional_event_writer_tests.rs
@@ -57,11 +57,12 @@ pub fn test_transactional_event_stream_writer(config: PravegaStandaloneServiceCo
     ));
 
     let mut writer =
-        handle.block_on(client_factory.create_transactional_event_writer(scoped_stream, WriterId(0)));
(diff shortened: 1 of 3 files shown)


integration_test/src/transactional_event_writer_tests.rs
```diff
@@ -174,6 +174,44 @@ async fn test_write_and_read_transaction(writer: &mut TransactionalEventWriter,
     info!("test write transaction passed");
 }
 
+async fn test_multiple_transactions(factory: &ClientFactory, stream: ScopedStream) {
```
shrids, line 177: Is this test being invoked?

shrids (changes requested):

integration_test/src/transactional_event_writer_tests.rs
Tristan1900: fixed, thanks!

shrids (approved):

tkaitchuck (approved):

PR #297, opened 2021-07-21
Issue 239: Add flush for event writer

### LINKED ISSUE #239: Document how to ensure durable write with event writer
**Problem description**
It is not clear how to use the event writer to determine when the event has been durably persisted.
Based on logging, `writer.write_event(payload).await` returns even before the event has been sent across the wire.

**Problem location**
`EventStreamWriter.write_event()` and
https://github.com/pravega/pravega-client-rust/blob/master/examples/event_write_and_read.rs#L65

The documentation for `EventStreamWriter.write_event()` should be updated. Also, the example in event_write_and_read.rs should show how to await a durable write ack.

Tristan1900:
**Change log description**  
Add flush for event writer.

**Purpose of the change**  
Fix #239 

**What the code does**  
Add a flush method for event writer. Applications use flush to make sure events are persisted in Pravega.

**How to verify it**  
Tests should pass

### DIFF
--- a/src/event/writer.rs
+++ b/src/event/writer.rs
@@ -21,6 +21,8 @@ use tokio::sync::oneshot;
 use tracing::info_span;
 use tracing_futures::Instrument;
 
+type EventHandle = oneshot::Receiver<Result<(), Error>>;
+
 /// Write events exactly once to a given stream.
 ///
 /// EventWriter spawns a `Reactor` that runs in the background for processing incoming events.
@@ -75,6 +77,7 @@ use tracing_futures::Instrument;
 pub struct EventWriter {
     writer_id: WriterId,
     sender: ChannelSender<Incoming>,
+    event_handle: Option<EventHandle>,
 }
 
 impl EventWriter {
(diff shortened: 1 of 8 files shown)


src/event/writer.rs
```diff
@@ -112,37 +116,51 @@ impl EventWriter {
     /// ```ignore
     /// let mut event_writer = client_factory.create_event_writer(stream);
     /// // result is a tokio oneshot
-    /// let result = event_writer.write_event(payload).await;
-    /// result.await.expect("flush to server");
+    /// event_writer.write_event(payload).await;
+    /// event_writer.flush().await.expect("flush to server");
     /// ```
-    pub async fn write_event(&mut self, event: Vec<u8>) -> oneshot::Receiver<Result<(), Error>> {
+    pub async fn write_event(&mut self, event: Vec<u8>) {
```
tkaitchuck, line 122: I don't think we can make this change.
Unlike a byteStream where if one write fails all those after it are sure to fail, and we can resume knowing the offset;
EventStreams have multiple segments, so one event can error while other events before and after can succeed. Similarly the order of acks may be different for different segments.

We can have a flush method, but it does not remove the need for the ack callbacks.

tkaitchuck (changes requested):

Tristan1900:
Close this as it does not fit anymore.

PR #303, opened 2021-08-05
Issue 301 : Use native roots for reqwest

### LINKED ISSUE #301: Reqwest tls support
**Problem description**
Reqwest config should be allowed to add certificate. 

**Problem location**
clientconfig

Tristan1900:
**Change log description**  
Use native roots for reqwest which will auto scan the default linux cert directory. Previously, it is using root certs from webpki, so any self signed certs cannot be loaded. 

**Purpose of the change**  
Fix #301

**What the code does**  
Use native roots

**How to verify it**  
Had verified that it's working in SDP

### DIFF
--- a/config/Cargo.toml
+++ b/config/Cargo.toml
@@ -20,7 +20,7 @@ derive_builder = "0.9"
 getset = "0.0.9"
 serde_json = "1.0"
 serde = "1.0"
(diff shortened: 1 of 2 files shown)


claudiofahey (approved):
I confirmed that this change works with Pravega, Keycloak, and TLS using private CA certificates.

tkaitchuck (approved):

PR #309, opened 2021-08-26
Issue 310 : Enable inflight event configuration on Python writer API.

### LINKED ISSUE #310: Performance improvements for Python Event Writer API
**Problem description**
At the moment Python API to write events into Pravega ensures only one inflight event at any given point in time. This causes a drop in the performance of the Python API.

**Problem location**
Python Event Writer

Ensure we can have multiple inflight events with the python writer APIs.

shrids:
**Change log description**  
- Enable a configurable number of inflight events to be written by the Python writer.

**Purpose of the change**  
Fixes #310 

**What the code does**  
- Create a oneshot receiver holder which tracks multiple oneshot receivers tracking every inflight event.
- Users will now be able to specify the max number of inflight events that can be configured for the Python writer.
```rust
  pub fn create_writer(
        &self,
        scope_name: &str,
        stream_name: &str,
        max_inflight_events: usize,
    ) -> PyResult<StreamWriter> {..}
```
- A flush () API has been enabled for the Python Writer which ensures all the inflight events are pesisted.

``` rust
///
/// Flush all the inflight events into Pravega Stream.
/// This will ensure all the inflight events are completely persisted on the Pravega Stream.
///
pub fn flush(&mut self) {...}
```

**How to verify it**  


src/lib.rs
```diff
@@ -55,6 +55,6 @@ pub(crate) mod segment;
 #[doc(hidden)]
 pub mod test_utils;
 #[macro_use]
-mod util;
+pub mod util;
```
Tristan1900, line 58: We probably want to hide this module in document. Like `#[doc(hidden)]`

python_binding/src/stream_writer.rs
```diff
@@ -137,12 +157,31 @@ impl StreamWriter {
...
+    ///
+    /// Flush all the inflight events into Pravega Stream.
+    /// This will ensure all the inflight events are completely persisted on the Pravega Stream.
+    ///
+    #[text_signature = "($self)"]
+    #[args("*")]
+    pub fn flush(&mut self) {
+        info!("Invoking flush() on writer {:?}", self.to_str());
+        for x in self.inflight.drain() {
+            let _res = self.factory.runtime().block_on(x);
```
Tristan1900, line 169: We are ignoring the oneshot result. What if the result is `Err`, I think we should bubble it up to the caller or throw an exception.

src/lib.rs
shrids: done.

python_binding/src/stream_writer.rs
shrids: yes, I have improved the code.

src/util/oneshot_holder.rs
```diff
@@ -0,0 +1,155 @@
...
+//
+
+use std::collections::vec_deque::Drain;
+use std::collections::VecDeque;
+use tokio::sync::oneshot::error::RecvError;
+use tokio::sync::oneshot::Receiver;
+
+pub struct OneShotHolder<E> {
+    size: usize,
+    inflight: VecDeque<tokio::sync::oneshot::Receiver<Result<(), E>>>,
```
tkaitchuck, line 18: Receiver is imported. No need to specify the full path.

tkaitchuck (approved):

src/util/oneshot_holder.rs
shrids: removed the full path.

Tristan1900 (approved):

PR #312, opened 2021-09-06
Issue 313: Enable ByteStreamAPI for PravegaPython bindings.

### LINKED ISSUE #313: Enable ByteStream bindings for Python.
**Problem description**
Enable ByteStream bindings for Python.

**Problem location**
Python bindings

Implement it.

shrids:
**Change log description**  
Enable ByteStream API for Pravega Python bindings.

**Purpose of the change**  
Fixes #313 

**What the code does**  
Create a Binary I/O representation for a Pravega Stream in Python bindings. This implements the APIs provided by [docs.python.org/3/library/io.html#io.IOBase](https://docs.python.org/3/library/io.html#io.IOBase)

```python
import pravega_client;
manager=pravega_client.StreamManager("tcp://127.0.0.1:9090")

# Create a writer against an already created Pravega scope and Stream.
byte_stream=manager.create_byte_stream("scope", "stream");

# write bytes into the Pravega Stream. It returns the current write offset
byte_stream.write(b"bytes")
5

# seek to the specified read offset.
byte_stream.seek(2, 0)

#  read data into a buffer, it returns the number of bytes read.
buf=bytearray(5)
byte_stream.readinto(buf)
3

# print the buffer.
buf
bytearray(b'tes\x00\x00')

# truncate a byte stream at offset 2
byte_stream.truncate(2)

# fetch the current head offset of the Stream
byte_stream.current_head_offset()
2

# fetch the current tail offset of the Pravega Stream.
byte_stream.current_tail_offset()
5
```

Note: This fixes the cross compilation build issue with `cross` in the Github action.

**How to verify it**  


tkaitchuck (approved):

python_binding/src/byte_stream.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// Whether the buffer is seekable; here just for compatibility, it always returns True.
+    ///
+    pub fn seekable(&self) -> PyResult<bool> {
+        Ok(true)
+    }
+
+    ///
+    /// Get the current read position in the Stream.
+    ///
+    pub fn tell(&self) -> PyResult<u64> {
```
Tristan1900, line 169: why do we name it `tell`, is there a naming convention in python?

python_binding/src/byte_stream.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// byte_stream.flush();
+    /// buf = bytearray(5)
+    /// byte_stream.readinto(buf)
+    /// buf
+    /// bytearray(b'bytes')
+    /// ```
+    ///
+    #[text_signature = "($self, byteArray)"]
+    #[args(byteArray)]
+    pub fn readinto(&mut self, buf: &PyByteArray) -> PyResult<usize> {
```
Tristan1900, line 119: Why do we name it `readinto` instead of `read`, same naming convention as `tell`?

python_binding/src/byte_stream.rs
shrids: This maps the Python API https://docs.python.org/3/library/io.html#io.IOBase.tell

python_binding/src/byte_stream.rs
shrids: Yeah, I followed the Python API pattern here. ref: https://docs.python.org/3/library/io.html#io.RawIOBase.readinto

Tristan1900 (approved):

PR #331, opened 2021-11-05
Issue 286: Support reconnecting when a connection was never established

### LINKED ISSUE #286: panic with message 'should always have connection here'
**Problem description**
thread 'tokio-runtime-worker' panicked at 'should always have connection here', /Users/wenquan.xing/.cargo/git/checkouts/pravega-client-rust-fa7c139c5174e088/8954c04/src/segment/writer.rs:409:13

**Problem location**
see above

tkaitchuck:
Signed-off-by: Tom Kaitchuck <tom.kaitchuck@emc.com>

**Change log description**  
Support reconnecting when a connection was never established

**Purpose of the change**  
Fixes #286

**What the code does**  
Allows for no connection to be specified in a re-connect request. Meaning the reconnect should be attempted if there is no connection.

### DIFF
--- a/src/segment/reactor.rs
+++ b/src/segment/reactor.rs
@@ -68,6 +68,25 @@ impl Reactor {
                 }
                 Ok(())
             }
+            Incoming::Reconnect(writer_info) => {
+                let option = selector.writers.get_mut(&writer_info.segment);
+                if option.is_none() {
+                    return Ok(());
+                }
+                let writer = option.unwrap();
+
+                // Only reconnect if the current connection is having connection failure. It might happen that the
+                // write op has already triggered the connection failure and has reconnected. It's necessary to avoid
+                // reconnect twice since resending duplicate inflight events will cause InvalidEventNumber error.
+                let current_connection_id = writer.connection.as_ref().map(|c| c.get_id());
+                if current_connection_id == writer_info.connection_id {
+                    warn!("reconnect for writer {:?}", writer_info);
+                    writer.reconnect(factory).await;
+                } else {
+                    info!("reconnect signal received for writer: {:?}, but does not match current writer: id {}, connection id {:?}, ignore", writer_info, writer.id, current_connection_id);
+                }
+                Ok(())
+            }
             Incoming::ServerReply(server_reply) => {
                 if let Err(e) = Reactor::process_server_reply(server_reply, selector, factory).await {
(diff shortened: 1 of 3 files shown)


src/segment/reactor.rs
```diff
@@ -68,36 +68,33 @@ impl Reactor {
...
-                        writer.reconnect(factory).await;
-                    } else {
-                        info!("reconnect signal received for writer: {:?}, but does not match current writer: id {}, connection id {}, ignore", writer_info, writer.id, write_half.get_id());
-                    }
+
+                // Only reconnect if the current connection is having connection failure. It might happen that the
+                // write op has already triggered the connection failure and has reconnected. It's necessary to avoid
+                // reconnect twice since resending duplicate inflight events will cause InvalidEventNumber error.
+                let current_connection_id = writer.connection.as_ref().map(|c| c.get_id());
+                if current_connection_id == writer_info.connection_id {
```
shrids, line 82: If current_connection_id  is `None` we would need to invoke reconnect here right?

tkaitchuck: Walking through all the cases:
* If current_connection_id is None and writer_info.connection_id is None - Then we need to invoke reconnect.
* If current_connection_id is not none and writer_info.connection_id is None - Then we don't need to invoke reconnect because it has already happened, and writer.connection is the new connection.
* If both current_connection_id and writer_info.connection_id are not None and are the same value - Then this indicates this connection has failed, and reconnect needs to be called.
* If both current_connection_id and writer_info.connection_id are not None and are different values - Then reconnect does not need to be called because the failure is about an old connection that has already been replaced.

This only leaves: 
* If current_connection_id is None and writer_info.connection_id is not None. - This shouldn't be possible. current_connection_id is None when no connection has been established. However writer_info.connection_id contains the ID of the connection that should be restarted. So this would imply that a specific connection is being asked to be restarted when there have never been any connections. 

If this were to somehow occur, the code as-written treats this as a stale reconnect request and does nothing. It is debatable as to what "should" happen given that this shouldn't occur. However it IS a stal event in that the connection clearly changed (to none) between the time the request was queued and executed. If this were to happen in a real application the next attempt to use the connection would result in another reconnect attempt being issued with 'none' set because there isn't yet any connection which would return the code to a normal working state.

shrids (approved):

Issue #340, opened 2021-11-17
Implement Large Event Support


dellThejas:
**Problem description**
For large events we need to add additional support

**Problem location**
writer/event_writer.rs

**Suggestions for an improvement**
Add a flush implementation and add use transient segments to handle large events

PR #341, opened 2021-11-17
Issue 340: Fix event writer implementation and add flush

dellThejas:
Signed-off-by: dellThejas <Thejas.Vidyasagar@dell.com>

**Change log description**  
Added flush for event writer
Updated implementation of Event Writer

**Purpose of the change**  
#340

**What the code does**  
Flush implementation added to event writer similar to Byte writer
Also, event writer will not be return the receiver anymore. This implementation helps setup the flush, and makes the write operation easier to implement for users.

**How to verify it**  
ran existing tests

### DIFF
--- a/src/event/writer.rs
+++ b/src/event/writer.rs
@@ -18,7 +18,9 @@ use pravega_client_channel::{create_channel, ChannelSender};
 use pravega_client_shared::{ScopedStream, WriterId};
 
 use std::collections::VecDeque;
+use tokio::io::ErrorKind;
 use tokio::sync::oneshot;
+use tokio::sync::oneshot::error::TryRecvError;
 use tracing::info_span;
 use tracing_futures::Instrument;
 
@@ -191,11 +193,18 @@ impl EventWriter {
     /// ```
     pub async fn flush(&mut self) -> Result<(), Error> {
         while self.event_handles.front().is_some() {
-            let handle = self.event_handles.pop_front().expect("get first handle");
-            let event_result = handle.await.map_err(|e| Error::InternalFailure {
-                msg: format!("oneshot error {:?}", e),
-            })?;
-            event_result?;
+            let mut receiver = self.event_handles.pop_front().expect("get first handle");
+
+            let event_res = receiver.try_recv();
+            match event_res {
+                Err(TryRecvError::Closed) => {}
+                Err(TryRecvError::Empty) => {}
+                _ => {
+                    event_res.map_err(|e| Error::InternalFailure {
+                        msg: format!("oneshot error {:?}", e),
+                    })?;
+                }
+            }
         }
         Ok(())
     }
@@ -211,6 +220,10 @@ impl EventWriter {
     /// assert!(event_writer.flush_cleared());
     ///   ```
     pub fn flush_cleared(&self) -> bool {
+        // for  self.event_handles.front().is_some() {
+        //
+        // }
+
         self.event_handles.is_empty()
     }
 }
--- a/src/event/writer.rs
+++ b/src/event/writer.rs
@@ -17,9 +17,12 @@ use crate::util::get_random_u128;
 use pravega_client_channel::{create_channel, ChannelSender};
 use pravega_client_shared::{ScopedStream, WriterId};
 
+use std::collections::VecDeque;
 use tokio::sync::oneshot;
 use tracing::info_span;
 use tracing_futures::Instrument;
+
+// type EventHandle = oneshot::Receiver<Result<(), Error>>;
 
 /// Write events exactly once to a given stream.
 ///
@@ -75,6 +78,7 @@ use tracing_futures::Instrument;
 pub struct EventWriter {
     writer_id: WriterId,
     sender: ChannelSender<Incoming>,
+    event_handles: VecDeque<oneshot::Receiver<Result<(), Error>>>,
 }
 
 impl EventWriter {
@@ -93,6 +97,7 @@ impl EventWriter {
         EventWriter {
             writer_id,
             sender: tx,
+            event_handles: VecDeque::new(),
         }
     }
 
@@ -118,9 +123,18 @@ impl EventWriter {
     pub async fn write_event(&mut self, event: Vec<u8>) -> oneshot::Receiver<Result<(), Error>> {
         let size = event.len();
         let (tx, rx) = oneshot::channel();
+        let (tx_flush, rx_flush) = oneshot::channel();
         let routing_info = RoutingInfo::RoutingKey(None);
+        let routing_info_flush = RoutingInfo::RoutingKey(None);
         if let Some(pending_event) = PendingEvent::with_header(routing_info, event, None, tx) {
             let append_event = Incoming::AppendEvent(pending_event);
+            if let Some(pending_event_flush) =
+                PendingEvent::with_header(routing_info_flush, Vec::new(), None, tx_flush)
+            {
+                let append_event_flush = Incoming::AppendEvent(pending_event_flush);
+                let flush_rec = self.writer_event_internal(append_event_flush, 0, rx_flush).await;
+                self.event_handles.push_back(flush_rec);
+            }
             self.writer_event_internal(append_event, size, rx).await
         } else {
             rx
@@ -137,9 +151,18 @@ impl EventWriter {
     ) -> oneshot::Receiver<Result<(), Error>> {
         let size = event.len();
         let (tx, rx) = oneshot::channel();
+        let (tx_flush, rx_flush) = oneshot::channel();
         let routing_info = RoutingInfo::RoutingKey(Some(routing_key));
+        let routing_info_flush = RoutingInfo::RoutingKey(None);
         if let Some(pending_event) = PendingEvent::with_header(routing_info, event, None, tx) {
             let append_event = Incoming::AppendEvent(pending_event);
+            if let Some(pending_event_flush) =
+                PendingEvent::with_header(routing_info_flush, Vec::new(), None, tx_flush)
+            {
+                let append_event_flush = Incoming::AppendEvent(pending_event_flush);
+                let flush_rec = self.writer_event_internal(append_event_flush, 0, rx_flush).await;
+                self.event_handles.push_back(flush_rec);
+            }
             self.writer_event_internal(append_event, size, rx).await
         } else {
             rx
@@ -163,6 +186,32 @@ impl EventWriter {
         } else {
             rx
         }
+    }
+
+    /// Flush data.
+    ///
+    /// It will wait until all pending appends have acknowledgment.
+    ///
+    /// # Examples
+    /// ```ignore
+    /// let mut byte_writer = client_factory.create_byte_writer(segment).await;
+    /// let payload = vec![0; 8];
+    /// let size = event_writer.write_event(&payload).await;
+    /// event_writer.flush().await;
+    /// ```
+    pub async fn flush(&mut self) -> Result<(), Error> {
+        while self.event_handles.front().is_some() {
+            let handle = self.event_handles.pop_front().expect("get first handle");
+            let event_result = handle.await.map_err(|e| Error::InternalFailure {
+                msg: format!("oneshot error {:?}", e),
+            })?;
+            event_result?;
+        }
+        Ok(())
+    }
+
+    pub fn flush_cleared(&self) -> bool {
+        self.event_handles.is_empty()
     }
 }
 
--- a/src/event/writer.rs
+++ b/src/event/writer.rs
@@ -128,6 +128,7 @@ impl EventWriter {
         let routing_info_flush = RoutingInfo::RoutingKey(None);
         if let Some(pending_event) = PendingEvent::with_header(routing_info, event, None, tx) {
             let append_event = Incoming::AppendEvent(pending_event);
+            let write_event = self.writer_event_internal(append_event, size, rx).await;
             if let Some(pending_event_flush) =
                 PendingEvent::without_header(routing_info_flush, Vec::new(), None, tx_flush)
             {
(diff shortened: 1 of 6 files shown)


src/event/writer.rs
```diff
@@ -118,9 +123,18 @@ impl EventWriter {
     pub async fn write_event(&mut self, event: Vec<u8>) -> oneshot::Receiver<Result<(), Error>> {
         let size = event.len();
         let (tx, rx) = oneshot::channel();
+        let (tx_flush, rx_flush) = oneshot::channel();
         let routing_info = RoutingInfo::RoutingKey(None);
+        let routing_info_flush = RoutingInfo::RoutingKey(None);
         if let Some(pending_event) = PendingEvent::with_header(routing_info, event, None, tx) {
             let append_event = Incoming::AppendEvent(pending_event);
+            if let Some(pending_event_flush) =
```
shrids, line 131: Should'nt we do this logic post writring append_event? As I understand with this logic `flush()` would still have one event outstanding, right?

dellThejas: I think youre right! Let me fix it.

src/event/writer.rs
```diff
@@ -137,10 +152,20 @@ impl EventWriter {
...
         let size = event.len();
         let (tx, rx) = oneshot::channel();
+        let (tx_flush, rx_flush) = oneshot::channel();
         let routing_info = RoutingInfo::RoutingKey(Some(routing_key));
+        let routing_info_flush = RoutingInfo::RoutingKey(None);
         if let Some(pending_event) = PendingEvent::with_header(routing_info, event, None, tx) {
             let append_event = Incoming::AppendEvent(pending_event);
-            self.writer_event_internal(append_event, size, rx).await
+            let write_event = self.writer_event_internal(append_event, size, rx).await;
+            if let Some(pending_event_flush) =
```
shrids, line 161: Should we handle the else part here?

src/event/writer.rs
```diff
@@ -164,6 +189,32 @@ impl EventWriter {
...
+            let handle = self.event_handles.pop_front().expect("get first handle");
+            let event_result = handle.await.map_err(|e| Error::InternalFailure {
+                msg: format!("oneshot error {:?}", e),
+            })?;
+            event_result?;
+        }
+        Ok(())
+    }
+
+    pub fn flush_cleared(&self) -> bool {
```
shrids, line 215: public API requires a doc.

shrids (commented):
The changes are looking good. I have a couple of comments/questions.

src/event/writer.rs
dellThejas: Added a panic statement

src/event/writer.rs
dellThejas: Added doc as well.

src/event/writer.rs
```diff
@@ -118,10 +123,22 @@ impl EventWriter {
...
-            self.writer_event_internal(append_event, size, rx).await
+            let write_event = self.writer_event_internal(append_event, size, rx).await;
+            if let Some(pending_event_flush) =
+                PendingEvent::without_header(routing_info_flush, Vec::new(), None, tx_flush)
+            {
+                let append_event_flush = Incoming::AppendEvent(pending_event_flush);
+                let flush_rec = self.writer_event_internal(append_event_flush, 0, rx_flush).await;
+                self.event_handles.push_back(flush_rec);
+            } else {
+                panic!("Failed to track event in flush.");
```
shrids, line 139: @tkaitchuck is there a better way to solve this ?

shrids (approved):

src/event/writer.rs
```diff
@@ -118,10 +123,22 @@ impl EventWriter {
     pub async fn write_event(&mut self, event: Vec<u8>) -> oneshot::Receiver<Result<(), Error>> {
         let size = event.len();
         let (tx, rx) = oneshot::channel();
+        let (tx_flush, rx_flush) = oneshot::channel();
         let routing_info = RoutingInfo::RoutingKey(None);
+        let routing_info_flush = RoutingInfo::RoutingKey(None);
         if let Some(pending_event) = PendingEvent::with_header(routing_info, event, None, tx) {
             let append_event = Incoming::AppendEvent(pending_event);
-            self.writer_event_internal(append_event, size, rx).await
+            let write_event = self.writer_event_internal(append_event, size, rx).await;
```
tkaitchuck, line 131: Instead of calling writer_event_internal here, it might be better to modify it to have a single call which takes two `rx`es. (IE embedding both `tx`es into the same PendingEvent.)

This would avoid a bug which is currently in this code which is that it is passing `None` as the routing key for both the `rx_flush` and the `rx` which implies that it is legal for these to be sent to different segments.

src/event/writer.rs
tkaitchuck: `.expect()`

src/event/writer.rs
```diff
@@ -137,10 +154,22 @@ impl EventWriter {
     ) -> oneshot::Receiver<Result<(), Error>> {
         let size = event.len();
         let (tx, rx) = oneshot::channel();
+        let (tx_flush, rx_flush) = oneshot::channel();
         let routing_info = RoutingInfo::RoutingKey(Some(routing_key));
+        let routing_info_flush = RoutingInfo::RoutingKey(None);
```
tkaitchuck, line 159: This will route the flush event differently from the event it is supposed to be following. This could result in flush returning even when the data has not yet been written. Please add a test that covers this case.

tkaitchuck (changes requested):

src/event/writer.rs
dellThejas: I added an additional optional Sender within PendingEvent, and write_event_internal now is only called once to track flush activity as well as write. This should eliminate our concerns dealing with different routing keys, and different segments.

src/event/writer.rs
```diff
@@ -17,10 +17,13 @@ use crate::util::get_random_u128;
 use pravega_client_channel::{create_channel, ChannelSender};
 use pravega_client_shared::{ScopedStream, WriterId};
 
+use std::collections::VecDeque;
 use tokio::sync::oneshot;
 use tracing::info_span;
 use tracing_futures::Instrument;
 
+// type EventHandle = oneshot::Receiver<Result<(), Error>>;
```
tkaitchuck, line 25: Either delete this or enable it.

src/event/writer.rs
```diff
@@ -93,6 +97,7 @@ impl EventWriter {
         EventWriter {
             writer_id,
             sender: tx,
+            event_handles: VecDeque::new(),
```
tkaitchuck, line 100: I don't see the completed handles being removed. It would be good to do this is write_event after the write to prevent this from growing forever in memory if the user never calls flush.

src/event/writer.rs
```diff
@@ -161,9 +173,46 @@ impl EventWriter {
...
+    /// # Examples
+    /// ```ignore
+    /// let mut byte_writer = client_factory.create_byte_writer(segment).await;
+    /// let payload = vec![0; 8];
+    /// let size = event_writer.write_event(&payload).await;
+    /// event_writer.flush().await;
+    /// assert!(event_writer.flush_cleared());
+    ///   ```
+    pub fn flush_cleared(&self) -> bool {
+        self.event_handles.is_empty()
```
tkaitchuck, line 214: This doesn't handle the case where the are event_handles, but they are completed.

src/event/writer.rs
```diff
@@ -161,9 +173,46 @@ impl EventWriter {
...
+    ///
+    /// # Examples
+    /// ```ignore
+    /// let mut byte_writer = client_factory.create_byte_writer(segment).await;
+    /// let payload = vec![0; 8];
+    /// let size = event_writer.write_event(&payload).await;
+    /// event_writer.flush().await;
+    /// assert!(event_writer.flush_cleared());
+    ///   ```
+    pub fn flush_cleared(&self) -> bool {
```
tkaitchuck, line 213: Can we have a better name? Maybe `is_flushed` or `has_inflight`?

src/event/writer.rs
dellThejas: I think it is better to just take this out, because the flush() is already checking all handles and trying to use receiver to tell if it has completed or not. The other option is to track the senders for each oneshot, which doesn't seem optimal.

src/event/writer.rs
dellThejas: Now, it will remove the receivers when completed.

src/event/writer.rs
dellThejas: Removed

src/event/writer.rs
```diff
@@ -161,9 +172,39 @@ impl EventWriter {
...
+    /// # Examples
+    /// ```ignore
+    /// let mut byte_writer = client_factory.create_byte_writer(segment).await;
+    /// let payload = vec![0; 8];
+    /// let size = event_writer.write_event(&payload).await;
+    /// event_writer.flush().await;
+    /// ```
+    pub async fn flush(&mut self) -> Result<(), Error> {
+        while self.event_handles.front().is_some() {
+            let mut receiver = self.event_handles.swap_remove_front(0).expect("get first handle");
```
tkaitchuck, line 193: I think `pop_front()` would be preferable to `swap_remove_front(0)` here.
This could also be combined with the conditional as in:
```
while let Some(receiver) = self.event_handles.front()
```
To be more efficient, clear, and concise.

src/event/writer.rs
```diff
@@ -161,9 +172,39 @@ impl EventWriter {
...
             rx
         }
     }
+
+    /// Flush data.
+    ///
+    /// It will wait until all pending appends have acknowledgment.
+    ///
+    /// # Examples
+    /// ```ignore
```
tkaitchuck, line 185: Remove the ignore. This example is broken.

src/event/writer.rs
```diff
@@ -161,9 +172,39 @@ impl EventWriter {
...
+    /// let mut byte_writer = client_factory.create_byte_writer(segment).await;
+    /// let payload = vec![0; 8];
+    /// let size = event_writer.write_event(&payload).await;
```
tkaitchuck, line 188: A size is not returned here.

src/event/writer.rs
```diff
@@ -161,9 +172,39 @@ impl EventWriter {
...
+            let mut receiver = self.event_handles.swap_remove_front(0).expect("get first handle");
+
+            let try_recv = receiver.try_recv();
+            match try_recv {
+                Err(TryRecvError::Closed) => {}
+                Err(TryRecvError::Empty) => {}
+                _ => {
+                    try_recv.map_err(|e| Error::InternalFailure {
+                        msg: format!("oneshot error {:?}", e),
+                    })??;
```
tkaitchuck, line 202: is there an await missing here?
It looks like this method will just return without actually waiting on anything.
Also instead of `_` it's probably worth explicitly having an `Ok` case as this branch is currently handling both all unlisted errors and the success case.

src/event/writer.rs
```diff
@@ -161,9 +172,39 @@ impl EventWriter {
                 .expect("send error");
             rx_error
         } else {
+            self.event_handles.push_back(rx_flush);
```
tkaitchuck, line 175: We need to support the case where the caller never calls flush without leaking memory. Currently the flush call removes items from `event_handles` but if it is not called this code will OOM.

tkaitchuck (changes requested):

src/event/writer.rs
dellThejas: Do we need to periodically call flush to clear this? Or upon every write, call flush on all previous writes? Or clean up event_handles while doing every write? (It would need to go through every write call to check if its processed or not)

src/segment/event.rs
```diff
@@ -50,6 +50,7 @@ pub(crate) struct PendingEvent {
     pub(crate) data: Vec<u8>,
     pub(crate) conditional_offset: Option<i64>,
     pub(crate) oneshot_sender: oneshot::Sender<Result<(), Error>>,
+    pub(crate) flush_oneshot_sender: Option<oneshot::Sender<Result<(), Error>>>,
```
shrids, line 53: flush_oneshot_sender is never used anywhere; we would need to invoke a send() with the write status on it.

shrids (changes requested):

src/event/writer.rs
dellThejas: Issue here was that I wasnt sending anything to tx_flush sender. Now, I am so we wont have this issue.

src/event/writer.rs
tkaitchuck: Yes. On a write call, clear out any completed ones stopping at the first uncompleted one.

src/event/writer.rs
```diff
@@ -161,9 +172,43 @@ impl EventWriter {
...
+    fn clear_initial_complete_events(&mut self) {
+        while let Some(mut receiver) = self.event_handles.pop_front() {
+            let try_recv = receiver.try_recv();
+
+            match try_recv {
+                Err(TryRecvError::Empty) => {
+                    self.event_handles.push_front(receiver);
+                    break;
+                }
+                Err(TryRecvError::Closed) => {}
```
shrids, line 207: I think we should return an error here and fail fast.

src/event/writer.rs
```diff
@@ -161,9 +172,43 @@ impl EventWriter {
...
+                _ => {}
```
tkaitchuck, line 208: Please make this:
```
Ok(_) => {}
```
That way if a new error value is ever added to `try_receive` we'll get a compile error rather than being unaware of it and simply skipping over the error at runtime.

src/segment/writer.rs
```diff
@@ -377,6 +377,14 @@ impl SegmentWriter {
                     acked.event_id
                 );
             }
+            if let Some(flush_sender) = acked.event.flush_oneshot_sender {
+                if flush_sender.send(Result::Ok(())).is_err() {
+                    trace!(
```
tkaitchuck, line 382: This should probably be info.

tkaitchuck (approved):

shrids (approved):
LGTM.

tkaitchuck (approved):

PR #342, opened 2021-12-02
Issue 319: Enable delete readergroup

### LINKED ISSUE #319: Python client should have a delete_reader_group function
**Problem description**
Python client has a create_reader_group function but there is no way to delete the reader group. This will clutter the scope with unused reader groups.

**Problem location**
Python client

Create a delete_reader_group function.

shrids:
**Change log description**  
- Enable delete ReaderGroup on the RUST and Python clients.

**Purpose of the change**  
Fixes #319 

**What the code does**  
- Enable delete ReaderGroup on the RUST and Python clients.

**How to verify it**  
TBD: Additional tests are being added.
All the existing tests and newly added tests should pass.


src/event/reader.rs
```diff
@@ -267,9 +267,23 @@ impl EventReader {
     /// is assumed dead.
     pub async fn release_segment(&mut self, mut slice: SegmentSlice) -> Result<(), EventReaderError> {
         info!(
-            "releasing segment slice {} from reader {}",
+            "releasing segment slice {} from reader {:?}",
             slice.meta.scoped_segment, self.id
         );
+        // check if the reader is already offline.
+        if self.meta.reader_offline || !self.rg_state.lock().await.check_online(&self.id).await {
```
tkaitchuck, line 274: This is going to add more traffic to the table. Can we combine this with the update by embedding the 'reader group offline' state into it implicitly? For example by detecting if the table is deleted.

src/event/reader.rs
```diff
@@ -322,6 +336,19 @@ impl EventReader {
             slice.meta.end_offset >= offset,
             "the offset where the segment slice is released should be less than the end offset"
         );
+        if self.meta.reader_offline || !self.rg_state.lock().await.check_online(&self.id).await {
```
tkaitchuck, line 339: see above

src/event/reader.rs
```diff
@@ -411,6 +438,19 @@ impl EventReader {
         mut slice: SegmentSlice,
         read_offset: i64,
     ) -> Result<(), EventReaderError> {
+        if self.meta.reader_offline || !self.rg_state.lock().await.check_online(&self.id).await {
```
tkaitchuck, line 441: see above

src/sync/table.rs
```diff
@@ -79,6 +81,76 @@ pub struct Table {
...
+            stream: PravegaStream::from(format!("{}{}", name, KVTABLE_SUFFIX)),
+            segment: Segment::from(0),
+        };
+        info!("deleting table map on {:?}", segment);
+
+        let delegation_token_provider = factory
+            .create_delegation_token_provider(ScopedStream::from(&segment))
+            .await;
+        let op = "Delete table segment";
+        retry_async(factory.config().retry_policy, || async {
```
tkaitchuck, line 103: A private function would help the indentation and wrapping here.

src/event/reader.rs
vangork: As per the comment of https://github.com/pravega/pravega-client-rust/blob/master/src/event/reader.rs#L360-L362, a reader could be brought offline by some other thread, should it be enough by just checking if the table is deleted? If yes, should we just send Requests::GetSegmentAttribute wirecommand to check if the table is deleted?

tkaitchuck (approved):

PR #344, opened 2021-12-08
Issue 343: Fix Byte Client sample code snips

### LINKED ISSUE #343: Sample code snippet is out of date
**Problem description**
The sync method current_head is already removed. The code snippet doesn't work any more.

**Problem location**
https://pravega.github.io/pravega-client-rust/Rust/ByteClient.html#read-from-an-offset
https://pravega.github.io/pravega-client-rust/Rust/ByteClient.html#put-it-together

Update the code snippets accordingly.

shrids:
**Change log description**  

* Fix the code snips

**Purpose of the change**  
FIxes #343

**What the code does**  

* Fix the code snips

**How to verify it**  
All the existing tests should continue to pass.

### DIFF
--- a/book/src/Rust/ByteClient.md
+++ b/book/src/Rust/ByteClient.md
@@ -77,7 +77,7 @@ then it reaches the end of segment and no more data could be read from this offs
 ```rust
 let offset = byte_reader.current_head().await.expect("get current head offset");
 let mut buf: Vec<u8> = vec![0; 4];
-let size = byte_reader.read(&mut buf).expect("read from byte stream");
+let size = byte_reader.read(&mut buf).await.expect("read from byte stream");
 ```
 
 ### Put it together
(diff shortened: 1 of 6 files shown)


tkaitchuck:
I don't feel like the handle method belongs on the byteReader api. It might be more appropriate on the ClientFactory,

shrids:
Yes makes sense, the ClientFactory already has a method to return the runtime handle.
```
pub fn runtime_handle(&self) -> Handle
```

The gstreamer plugin wraps a ByteReader inside a `std::io::BufReader`  Since the ByteReader now only has async APIs the idea was to enable a way to obtain the Runtime Handle from the ByteReader. 
As a next step the gstreamer plugin can pass a wrapper object which provides the sync/blocking APIs using the Handle returned by the ClientFactory.

tkaitchuck (approved):

PR #355, opened 2022-03-04
Issue 354: Nodejs StreamManager

thekingofcity:
Signed-off-by: thekingofcity <3353040+thekingofcity@users.noreply.github.com>

**Change log description**  
Add `StreamManager` support to Nodejs bindings.

**Purpose of the change**  
Fixes #354 

**What the code does**  
Use `neon-bindings` to wrap the rust client.

**How to verify it**  
1. `npm i`
2. `npm run build-debug`
3. `node --loader ts-node/esm tests/stream_manager.ts` should pass in a clean Pravega standalone deployment.

PR #358, opened 2022-03-11
Issue 357: Nodejs ReaderGroup and EventReader

### LINKED ISSUE #357: Nodejs ReaderGroup and EventReader
**Problem description**
Add `ReaderGroup` and `EventReader` support to Nodejs.

**Problem location**
`./nodejs/`

thekingofcity:
Signed-off-by: thekingofcity <3353040+thekingofcity@users.noreply.github.com>

**Change log description**  
Add `ReaderGroup` and `EventReader` support to Nodejs bindings.

**Purpose of the change**  
Fixes #357 

**What the code does**  
Implement the `stream_reader_group` and `stream_reader` just like the Python code.

It also uses the asynchronous approach so that users can do the followings:
```javascript
const seg_slice: Slice = await stream_reader.get_segment_slice();
const enc = new TextDecoder('utf-8');
for (const event of seg_slice) {
    const raw_value: ArrayBuffer = event.data();
    console.log(`Event at ${event.offset()} reads ${enc.decode(raw_value)}`);
}
```

This should be merged after #355.

Reference about `@@iterator`:
1. https://www.geekabyte.io/2019/06/typing-iterables-and-iterators-with.html
2. https://basarat.gitbook.io/typescript/future-javascript/iterators
3. https://stackoverflow.com/questions/52326671/the-syntax-of-symbol-iterator-is-difficult
4. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

**How to verify it**  
Since the writer is not here, I use Python code to create the stream and write something to Pravega.

1. `npm run build-debug`
2. `python3 tests/write.py` will write 4 events.
  ```
  b'e1'
  b'e2'
  b'e3'
  b'e4'
  ```
3. `node --loader ts-node/esm tests/stream_reader.ts` should print these.
  ```
  (node:5239) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
  (Use `node --trace-warnings ...` to show where the warning was created)
  Event at 0 reads e1
  Event at 10 reads e2
  Event at 20 reads e3
  Event at 30 reads e4
  ```

I will write similar tests like Python after the writer is implemented.


nodejs/src/stream_reader.rs
```diff
@@ -0,0 +1,209 @@
...
+        let event_data = cx.this().downcast_or_throw::<JsBox<EventData>, _>(&mut cx)?;
+        let data = event_data.value.as_slice().clone();
+
+        Ok(JsArrayBuffer::external(&mut cx, data.to_owned()))
+    }
+
+    ///
+    /// Return the event offset in the segment.
+    ///
+    pub fn js_offset(mut cx: FunctionContext) -> JsResult<JsNumber> {
```
tkaitchuck, line 50: Do we want to expose this?

nodejs/src/stream_reader.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    pub fn js_next(mut cx: FunctionContext) -> JsResult<JsBox<EventData>> {
+        // Use RefCell as self.seg_slice needs to be changed.
+        // Example at https://github.com/neon-bindings/neon/blob/0.10.0/test/napi/src/js/boxed.rs
+        let slice = cx.argument::<JsBox<RefCell<Slice>>>(0)?;
+        let event_data = slice.borrow_mut().next();
+
+        match event_data {
+            Some(event_data) => Ok(cx.boxed(event_data)),
+            // TODO: better return something like undefined instead of throwing an error?
```
tkaitchuck, line 103: What's the standard TS patten here? Throwing seems like an odd one. Does it typically have a `hasNext` like java? If so we could do that by a peek.

nodejs/src/stream_reader_group.rs
```diff
@@ -0,0 +1,144 @@
...
+    ///
+    fn reader_offline(&self, reader_name: &str) -> Result<(), ReaderGroupStateError> {
+        info!(
+            "Marking reader {:?} under reader group {:?} as offline",
+            reader_name, self.reader_group.name
+        );
+
+        let res = self
+            .runtime_handle
+            .block_on(self.reader_group.reader_offline(reader_name.to_string(), None));
```
tkaitchuck, line 94: Here it's passing None. Do we want to allow for passing an argument here?

nodejs/tests/stream_reader.ts
```diff
@@ -0,0 +1,33 @@
...
+
+try {
+    const seg_slice = await stream_reader.get_segment_slice();
+    const dec = new TextDecoder('utf-8');
+    for (const event of seg_slice) {
+        const raw_value = event.data();
+        console.log(`Event at ${event.offset()} reads ${dec.decode(raw_value)}`);
+    }
+} catch (e) {
+    console.log(e);
```
tkaitchuck, line 30: Is this going to work with things like github actions? We want to make sure that tests properly fail the build not just log the errors.

nodejs/src/stream_reader.rs
thekingofcity: Not sure. Will exposing this cause confusion to users? The Python `__repr__` implementation of `EventData` does expose the offset, so I'm assuming it's something that is ok to be exposed.

nodejs/src/stream_reader.rs
thekingofcity: I did some research on the `Iteration protocols`. Looks like as long as we implement a `next(): IteratorResult`, we should be good to go. It's up to us to decide the internal implementation (whether throwing an error or returning some custom-defined end thing).

Would you like to have an `undefined` pattern? It should looks like this:

```javascript
            const internal_event = SliceNext(slice); // <- It is hard to type internal_event as it is a external(rust) object here.
            return {
                done: internal_event ? true : false,
                value: internal_event ? Event(internal_event) : undefined,
            };
```

We could also implement a `hasNext` function, but that requires more code on rust side to determine the status and will leave nodejs and python binding different.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols
https://medium.com/bytelimes/iterables-in-javascript-custom-iterables-97a1f15852eb

nodejs/src/stream_reader_group.rs
thekingofcity: I'm doing the same thing as Python here. How about having an implementation reference in Python?😂 I'll update the `reader_offline` later. Same as #319

nodejs/tests/stream_reader.ts
thekingofcity: No, it will not. Even tests got rewritten in #361, they are not meant to be run in github actions. If we need to run a standalone Pravega in actions, we should create a new issue.

The catch thing is to make sure other readers would be created correctly if this fails as we have a single standalone running when testing. But I add the random names so they should be removed. I want to do it in #361 since this file is removed entirely in the next pr.

nodejs/src/stream_reader.rs
tkaitchuck: I am not worried about the API being different in different languages. As generally you are only working with one. So we should do for NodeJs what is most natural in NodeJs regardless of Python.

nodejs/tests/stream_reader.ts
tkaitchuck: We definitely want to be able to have tests run as a pre-submit step. Otherwise they are likely to simply be not run. 
We currently separate unit tests and integration tests which run separately where unit tests are assumed to be parallelism safe and integration tests are not. If this is using the standalone, then it's definitely an integration test. So we should just make sure to include it in that pre-submit step and that any errors will cause it to fail.

tkaitchuck (commented):
Clippy has some comments also

thekingofcity:
Ci tests and clippy problems are fixed.

nodejs/src/stream_reader.rs
thekingofcity: I would say the current implementation is sufficient since we write the right `next()`. There is no `hasNext` pattern in the js world.

nodejs/src/stream_reader.rs
tkaitchuck: It's fine to pass it to the application and for them to pass it back to us via APIs. But we want to be clear that one cann't do pointer math with it. IE: It is not ok to generate your own integer based on this value and then use that to refer to something, because we don't guarantee that we won't have any header or compression or something in the future. So it's fine so long as the value is not interpreted. 

In Java we do this by having an Object which wraps it which is opaque to users, so while it is just a long they can't directly get that long for visibility reasons. This is the normal pattern in Java to lock things down. In Python that is not the normal pattern and underscores are used in the name to indicate caution. I don't know what the norm is in JS. Is there a common pattern around this?

nodejs/src/stream_manager.rs
```diff
@@ -0,0 +1,632 @@
...
+        let scope_name = cx.argument::<JsString>(0)?.value(&mut cx);
+        let stream_name = cx.argument::<JsString>(1)?.value(&mut cx);
+
+        let tags = match stream_manager.get_stream_tags(&scope_name, &stream_name) {
+            Ok(val) => match val {
+                Some(val) => val,
+                None => return Ok(cx.empty_array()),
+            },
+            Err(_err) => {
+                // TODO: display the error details.
```
tkaitchuck, line 538: Please implement. (The display trait should provide a coherent message)

nodejs/src/stream_reader.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl StreamReader {
+    ///
+    /// Return a Slice in an asynchronous call.
+    /// The actural returned type from await will be `Promise<Slice>` aka `JsResult<JsBoxed<RefCell<Slice>>>`.
+    ///
+    /// See the tokio-fetch example for more details on how to return a Promise and await.
+    /// https://github.com/neon-bindings/examples/tree/2dbbef55f483635d0118c20c9902bf4c6faa1ecc/examples/tokio-fetch
+    ///
```
tkaitchuck, line 140: Is this meant to be a user facing comment?

nodejs/src/stream_reader_group.rs
tkaitchuck: I opened an issue here: https://github.com/pravega/pravega-client-rust/issues/370

tkaitchuck (approved):
Approved with comments

nodejs/src/stream_manager.rs
thekingofcity: display err with `format!`

nodejs/src/stream_reader.rs
thekingofcity: No, all comments in rust should not be a user facing comment/doc. Users are advised to read the README.md for a tutorial and https://pravega.github.io/pravega-client-rust/nodejs/index.html for a full API reference(auto generated from ts files and contains user facing comments.) The API reference is added in https://github.com/pravega/pravega-client-rust/pull/366/files#diff-d7ecc9db1d2f1f75440ba2c3e433aec8c252e9c86a6260732400737857c71d3b and should be available in the github.io pages.

nodejs/src/stream_reader.rs
thekingofcity: There is a keyword called `readonly` that works exactly as Python's underscore pattern. But if you've written Python or JavaScript or any similar dynamic code, you would know users could access and manipulate anything they want. So instead of posing some restrictions on offset, it may be better to state it in doc or comment and let users be clear about what they are doing.

Besides, both py and js bindings don't allow users to specify a streamcut/offset on reading, so we may also improve this later.

tkaitchuck (approved):

PR #367, opened 2022-04-11
Properly implement the retry policy

### LINKED ISSUE #359: Separate multiple from initial backoff in retry_policy
**Problem description**
Currently in the constructor `from_millis` the passed value is used for both the base value and the multiple. These should be separate parameters.

**Problem location**
RetryWithBackoff::from_millis

Take two arguments.

wxing1292:
**Change log description**  
* Re-implement the retry policy
  * Add initial delay
  * Add max delay
  * Add max attempt
  * Add expiration time
  * Add tests

**Purpose of the change**  
Fixes #359

### DIFF
--- a/controller-client/src/lib.rs
+++ b/controller-client/src/lib.rs
@@ -424,6 +424,7 @@ async fn get_channel(config: &ClientConfig) -> Channel {
     async { Channel::balance_list(endpoints.into_iter()) }.await
 }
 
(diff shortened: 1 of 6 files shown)


tkaitchuck (approved):

tkaitchuck:
Clippy is not happy. I think you'll have to add an "allow deprecated" annotation on the test class so that it works without error.

Issue #346, opened 2021-12-16
Create a Golang client init directory structure.


shrids:
**Problem description**
Create a Golang client init directory structure.

**Problem location**
Rust Project.

**Suggestions for an improvement**
Create a init directory structure for the current POC.

PR #377, opened 2022-06-13
Rust binding for Golang Client

vangork:
**Change log description**  
Create a rust binding for Golang client POC

**Purpose of the change**  
Continue #346

**What the code does**  
Create a rust binding for Golang client POC

**How to verify it**  
```
cd golang
go build examples/go/main.go
./main
```


golang/src/memory.rs
```diff
@@ -0,0 +1,69 @@
...
+pub extern "C" fn free_buffer(buf: Buffer) {
+    unsafe {
+        let _ = buf.consume();
+    }
+}
+
+impl Buffer {
+    // read provides a reference to the included data to be parsed or copied elsewhere
+    // data is only guaranteed to live as long as the Buffer
+    // (or the scope of the extern "C" call it came from)
```
tkaitchuck, line 23: This does not describe a safe function. I think this will have to be marked as unsafe. 
We may want to think about a way to reduce the unsafe surface. IE maybe doing some copying up front so that the bulk of the code can be safe.

golang/src/memory.rs
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    /// Creates an owned copy that can safely be stored and mutated.
+    #[allow(dead_code)]
+    pub fn to_owned(&self) -> Option<Vec<u8>> {
+        self.read().map(|slice| slice.to_owned())
+    }
+
+    // consume must only be used on memory previously released by from_vec
```
tkaitchuck, line 38: I don't understand this statment.

golang/src/memory.rs
```diff
@@ -0,0 +1,69 @@
...
+    // when the Vec is out of scope, it will deallocate the memory previously referenced by Buffer
+    pub unsafe fn consume(self) -> Vec<u8> {
```
tkaitchuck, line 40: Perhaps this should be called `to_vec`

golang/src/memory.rs
```diff
@@ -0,0 +1,69 @@
...
+        if self.is_empty() {
+            return Vec::new();
+        }
+        let mut v = Vec::from_raw_parts(self.ptr, self.len, self.cap);
+        v.shrink_to_fit();
+        v
+    }
+
+    // this releases our memory to the caller
```
tkaitchuck, line 49: This doc is assuming some odd context. This method doesn't release any memory...

golang/src/reactor.rs
```diff
@@ -0,0 +1,56 @@
...
+        receiver: &mut UnboundedReceiver<Incoming>,
+    ) -> Result<(), String> {
+        let incoming = receiver.recv().await.expect("sender closed, processor exit");
+        let chan_id = incoming.chan_id;
+        match incoming.operation {
+            Operation::GetSegmentSlice(reader) => {
+                let stream_reader = reader.get();
+                match stream_reader.get_segment_slice().await {
+                    Ok(seg_slice) => {
+                        let ptr = Box::into_raw(Box::new(Slice::new(seg_slice)));
```
tkaitchuck, line 38: I am not clear on where / when the segment slice is expected to be freed.

golang/src/stream_manager.rs
```diff
@@ -0,0 +1,354 @@
...
+            .controller_uri(controller_uri)
+            .is_auth_enabled(false)
+            .is_tls_enabled(false);
+        let config = builder.build().expect("creating config");
+        let client_factory = ClientFactory::new(config.clone());
+
+        // start reactor
+        let handle = client_factory.runtime_handle();
+        let (tx, rx) = unbounded_channel();
+        handle.spawn(Reactor::run(rx));
```
tkaitchuck, line 39: Humm.... I don't think this is what we want. 
Part of the idea was that if we have a go thread we can use it to call back into go for free. Here it is spawning a task on the generic background threadpool so it won't use the go thread. It would be better if we had a `run` method or something that go would call from it's wrapper and then we could capture the thread.

golang/src/stream_reader.rs
```diff
@@ -0,0 +1,167 @@
...
+impl SafeStreamReader {
+    pub fn new(v: *mut StreamReader) -> Self {
+        Self { v }
+    }
+    pub fn get(&self) -> &mut StreamReader {
+        unsafe { &mut *(self.v) }
+    }
+}
+
+unsafe impl Send for SafeStreamReader {}
```
tkaitchuck, line 52: As discussed, this is not safe. The compiler was correct to flag the error. We may need to either use an ID and a map, or find a way to pass ownership of a boxed reader. (Whichever ends up being easier)

tkaitchuck (commented):
One other thought:
A lot of the rules around cgo focus on not having the native code ever hold a pointer to go ram because the GC can move the object and the result will be data corruption. We could get the rust compiler to enforce these rules for us by immediately converting the pointer into a borrowed reference.

golang/src/config.rs
```diff
@@ -0,0 +1,298 @@
...
+use std::time::Duration;
+use pravega_client_config::credentials::Credentials;
+use pravega_client_retry::retry_policy::RetryWithBackoff;
+
+
+const SPLIT: &str = ",";
+
+
+#[repr(C)]
+pub struct BStreamConfiguration {
```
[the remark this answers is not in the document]
CharlieLChen: I rename all the struct names with the `Mapping` suffix, maybe it makes more sense. so the logic will be `Go.StreamConfiguration -> C.StreamConfigurationMapping -> rust.StreamConfiguration`. I think it is quite clear for type conversion between go and rust.
`StreamConfigurationMapping` is also distinguished from `StreamConfiguration`, so people will not get confused when they review the binding-related codes.

golang/src/stream_manager.rs
vangork: @tkaitchuck, I removed the reactor because the operations shouldn't be handled in one task which will serialize the operations. Also it won't need to retrieve identifiers from a Rust channel at Golang side as `ackOperationDone` would notify the Golang channel when the resource is ready at the Rust side.  I reimplement get_segment_slice as https://github.com/vangork/pravega-client-rust/blob/c97f91e077639bf56f8969cc572596155d721c47/golang/src/stream_reader.rs#L39-L52. Would like to have your comment again.

golang/src/stream_reader.rs
vangork: If my latest implement is accepted, this part won't be needed.

golang/bench_test/main.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+func main() {
+	f, _ := os.Create("profile")
+	pprof.StartCPUProfile(f)
+	defer pprof.StopCPUProfile()
+	url := flag.String("uri", "127.0.0.1:9090", "controller uri")
+	scope := flag.String("scope", "foo", "scope")
+	stream := flag.String("stream", "bar", "stream")
+	size := flag.Int("size", 1024 * 1024, "event size")
```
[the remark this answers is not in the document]
vangork: set to 1024.

golang/examples/c++/main.cc
```diff
@@ -0,0 +1,24 @@
+#include <iostream>
+extern "C"
+{
+    #include <pravega_client.h>
+}
+
+int main() {
```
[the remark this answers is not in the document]
vangork: removed.

golang/pkg/memory.go
```diff
@@ -0,0 +1,72 @@
...
+	cu8_ptr = *C.uint8_t
+	usize   = C.uintptr_t
+	cusize  = C.size_t
+	cint    = C.int
+	ci32    = C.int32_t
+	ci64    = C.int64_t
+	cbool   = C.bool
+)
+
+func receiveSlice(b C.Buffer) []byte {
```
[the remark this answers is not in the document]
vangork: rename to copyAndDestroyBuffer

golang/src/memory.rs
vangork: done

golang/src/reactor.rs
vangork: User has to call `defer slice.Close()`. to free the segment slice.

golang/src/memory.rs
vangork: removed

golang/pkg/reactor.go
```diff
@@ -0,0 +1,61 @@
...
+	bridgeChannel = make(chan Bridge)
+
+	go func() {
+		for {
+			select {
+			case bridge := <- bridgeChannel:
+				{
+					value, loaded := channelMap.LoadAndDelete(bridge.ChanId)
+					if !loaded {
+						fmt.Printf("unexpect channelId: %d", bridge.ChanId)
```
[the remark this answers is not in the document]
vangork: add closeReactor.

golang/pkg/reactor.go
```diff
@@ -0,0 +1,61 @@
...
+func publishBridge(chanId int64, objPtr uintptr, errorMsg *C.char) {
+	var bridge Bridge = Bridge {
+		ChanId: chanId,
+		ObjPtr: unsafe.Pointer(objPtr),
+	}
+	bridgeChannel <- bridge
+}
+
+func RunReactor() {
+	bridgeChannel = make(chan Bridge)
```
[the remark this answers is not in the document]
vangork: done.

golang/pkg/stream_reader.go
```diff
@@ -0,0 +1,59 @@
...
+	ptr := <- channel
+	slice := (*C.Slice)(ptr)
+
+	return &SegmentSlice{
+		Slice: slice,
+	}, nil
+}
+
+func (reader *StreamReader) Read() {
+}
```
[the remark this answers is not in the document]
vangork: removed.

golang/src/config.rs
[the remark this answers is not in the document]
CharlieLChen: I fixed some of them. The rest of them are something like "variant is never constructed: `Keycloak`", which means it's never used in this rust project, but it's necessary for generating the C types so that the golang can use it.

golang/pkg/stream_writer.go
```diff
@@ -0,0 +1,41 @@
...
+	Writer *C.StreamWriter
+}
+
+func (writer *StreamWriter) Close() {
+	C.stream_writer_destroy(writer.Writer)
+}
+
+func (writer *StreamWriter) WriteEventByRoutingKey(routingKey string, event []byte) error {
+	buf := C.Buffer{}
+	defer runtime.KeepAlive(event)
```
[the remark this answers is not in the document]
vangork: stream_writer_write_event will just send the event to the reactor channel which shouldn't last long , Do you think we can keep as what it is right now?

CharlieLChen:
@tkaitchuck There is no more space for CI builds
```
error: could not rename component file from '/home/runner/.rustup/tmp/qef9qnwargfmhqmk_dir/bk' to '/home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share'
error: could not rename component file from '/home/runner/.rustup/tmp/lcyugvdusqtnlgeg_dir/bk' to '/home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc'
error: failed to extract package (perhaps you ran out of disk space?): No space left on device (os error 28)
```

CharlieLChen:
@vangork I think we are good to make this PR ready :)

golang/pkg/stream_reader.go
```diff
@@ -0,0 +1,77 @@
...
+	ptr := <-channel
+	slice := (*C.Slice)(ptr)
+
+	return &SegmentSlice{
+		Slice: slice,
+	}, nil
+}
+
+// SegmentSlice represent the piece of a segment which contains multiple events and user can read it.
+/*
```
tkaitchuck, line 35: What is this?

golang/src/stream_reader.rs
```diff
@@ -0,0 +1,130 @@
...
+    handle.spawn(async move {
+        match stream_reader.get_segment_slice().await {
+            Ok(seg_slice) => {
+                let ptr = Box::into_raw(Box::new(Slice::new(seg_slice)));
+                unsafe {
+                    ackOperationDone(id, ptr as usize);
+                };
+            }
+            Err(err) => {
+                // TODO: send error msg through the channel
```
tkaitchuck, line 46: Please fix

golang/README.md
```diff
@@ -6,8 +6,32 @@ Pravega is an open source distributed storage service implementing Streams. It o
...
+2. import to your project with `go get`
+```
+go get github.com/pravega/pravega-client-rust/golang@latest
+``` 
+or you can replace the dependence with local version
+```
+go mod edit -replace github.com/pravega/pravega-client-rust/golang=/root/go/src/your_golang_code_path
+```
+
+3. Write your application with the reference to example code or bench_test code. As for how to set the clientConfig and streamConfig, you can read rust document as reference [ClientConfig](https://docs.rs/pravega-client-config/latest/pravega_client_config/struct.ClientConfig.html),[StreamConfig](https://github.com/pravega/pravega-client-rust/blob/master/shared/src/lib.rs#L508-L514)
```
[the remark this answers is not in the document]
a6dulaleem: added Step by step use of the goclient apis

golang/pkg/reactor.go
```diff
@@ -0,0 +1,80 @@
...
+var operationMap sync.Map
+
+var operationIdCounter int64 = 0
+
+// Register current request, will get response when rust side call ackOperationDone later.
+func registerOperation() (int64, chan Operation) {
+	channel := make(chan Operation)
+
+	// register the id and channel
+	// TODO: handle id overflow
```
tkaitchuck, line 21: You don't need to worry about 64 bits overflowing

golang/pkg/stream_reader.go
```diff
@@ -0,0 +1,75 @@
...
+			}, nil
+		} else if op.ErrPtr != nil {
+			errLen := C.int(op.ErrLen)
+			errPtr := (*C.char)(op.ErrPtr)
+			errMsg := C.GoStringN(errPtr, errLen)
+			return nil, errors.New(errMsg)
+		} else {
+			return nil, errors.New("Unknown errors")
+		}
+	case <-time.After(5 * time.Second):
```
tkaitchuck, line 48: We probably don't want to hardcode this time. Maybe it should be an argument to the function

golang/src/config.rs
```diff
@@ -0,0 +1,238 @@
...
+            retention: self.retention.to_retention(),
+            tags: str_to_tags(self.tags),
+        }
+    }
+}
+
+fn to_str<'a>(p_str: *const c_char) -> &'a str {
+    let raw = unsafe { CStr::from_ptr(p_str) };
+    raw.to_str()
+        .expect("Error: Failed to convert C string to Rust string.")
```
tkaitchuck, line 38: The most probable reason is that the passed value is not valid utf-8. This should be mentioned in the error message.

golang/src/config.rs
```diff
@@ -0,0 +1,238 @@
...
+pub struct RetryWithBackoffMapping {
+    initial_delay: u64,
+    backoff_coefficient: u32,
+    max_delay: u64,
+    max_attempt: i32,
+    expiration_time: i64,
+}
+
+impl RetryWithBackoffMapping {
+    pub unsafe fn to_retry_with_backoff(&self) -> RetryWithBackoff {
```
tkaitchuck, line 186: Why is this unsafe?
Can we narrow the scope?

golang/src/stream_writer.rs
```diff
@@ -0,0 +1,128 @@
...
+            runtime_handle,
+        }
+    }
+
+    pub async fn run_reactor(mut receiver: UnboundedReceiver<Incoming>) {
+        while StreamWriter::run_once(&mut receiver).await.is_ok() {}
+    }
+
+    async fn run_once(receiver: &mut UnboundedReceiver<Incoming>) -> Result<(), String> {
+        let incoming = receiver.recv().await.expect("sender closed, processor exit");
```
tkaitchuck, line 45: Do we want `expect` or `?` here?

tkaitchuck (approved):
Once the above comments are addressed, this should be good to go.

PR #382, opened 2022-07-18
Issue 371: Nodejs async list scopes and streams

### LINKED ISSUE #371: Nodejs async list scopes and streams
**Problem description**
Async operation on stream manager would be great as some operations may take a long time to proceed. Also, as there is no `ping` alike method in rust client, these async could be used to determine the aliveness of the Pravega connection.

**Problem location**
nodejs/stream_manager

thekingofcity:
**Change log description**  
Leverage rust async operations of the controller and list scopes and streams asynchronously.

Special thanks to @vangork who managed to use the controller inside an async future.
Special thanks to Mathieu Letendre-Jauniaux who shared this solution with me.
https://rust-bindings.slack.com/archives/C0H978T55/p1652026800964259?thread_ts=1651799942.621949&cid=C0H978T55

**Purpose of the change**  
Fix #371 

**What the code does**  
Call the rust function in tokio thread with a timeout. So the time of controller operations could be controlled even if there is no server.

**How to verify it**  
The following script won't hang when the server goes down and will log timeout on each second.

```typescript
const stream_manager = StreamManager('tcp://127.0.0.1:9090', false, false, true);

setInterval(async () => {
    stream_manager
        .list_scopes()
        .then(scopes => console.log(`${new Date()} ${scopes}`))
        .catch(() => console.log(`${new Date()} timeout`));
}, 1000);

const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
await delay(10000);
```


tkaitchuck:
Please fix the clippy error

nodejs/src/stream_manager.rs
```diff
@@ -450,23 +410,47 @@ impl StreamManager {
...
+        let (deferred, promise) = cx.promise();
+
+        // spawn an `async` task on the tokio runtime.
+        use std::sync::Arc;
+        let cf = Arc::new(stream_manager.cf.to_async());
+        let handle = stream_manager.cf.runtime_handle();
+        // enter the runtime context and call spawn with timeout
+        let _rt = handle.enter();
+        tokio::spawn(time::timeout(
+            Duration::from_millis(TIMEOUT_IN_MILLISECONDS),
```
tkaitchuck, line 426: I don't think it's our job to put a timeout here. If the user wants to limit their blocking they can do that in their own code. Especially as 1s is just totally arbitrary.

nodejs/src/stream_manager.rs
```diff
@@ -594,24 +578,54 @@ impl StreamManager {
...
```
tkaitchuck, line 595: Same here

tkaitchuck (changes requested):

thekingofcity:
Clippy is complaining about things other than nodejs. And I confirmed that the code is the same as the lastest master.

```
error: this let-binding has unit value
Error:    --> src/event/reader_group_state.rs:261:9
```

nodejs/src/stream_manager.rs
thekingofcity: If users want to limit they can only do it in nodejs. And in nodejs, you just neglect the result of the promise after some timeout. But by doing this way, the internal connection is still open and may or may not be timeout after some uncontrolled time periods.

With `Duration::from_millis(TIMEOUT_IN_MILLISECONDS)`, it is certain that there will be no more internal connection left unmanaged even if the user doesn't care about it anymore. And this timeout is similar to the python wrapper `const TIMEOUT_IN_SECONDS: u64 = 120;` where the limit is hardcoded.

But I agree we could use other options to control the actual timeout instead of a hardcoded 1 second.

tkaitchuck:
> Clippy is complaining about things other than nodejs. And I confirmed that this par is the same as the latest master.

Yes. Clippy continues to get smarter and break our code every time it finds a new error. We just need to fix such things as they come up.

nodejs/src/stream_manager.rs
thekingofcity: It should work as a sanity check. I updated the time to 120 sec.

tkaitchuck (approved):

PR #459, opened 2023-11-16
Making index record size configurable

### LINKED ISSUE #461: Make index record size configurable
**Problem description**
Current index record is fixed length of 4k bytes. It would waste quite some space if real index is small. It is better to be adapted to the real size when reading/writing.

**Problem location**
https://github.com/pravega/pravega-client-rust/blob/c0f18135798cf4f54512e13066d94aed1a71dfb8/src/index/mod.rs#L129

Default implementation for Record Size can be given in the Fields trait the same can be overridden if a different value is required.

AnkurAnand11:
**Change log description**  
Make index record size configurable. Current index record is of fixed length of 4k bytes. It would waste quite some space if real index is small. It is better to be adapted to the real size when reading/writing.

**Purpose of the change**  
Closes #461 

**What the code does**  
Created a new method get_record_size in the fields trait of mod.rs with default implementation of 4k bytes. If a different record size value is required, the traits method can be overridden while writing the index event.

**How to verify it**  
All test cases should pass.


src/index/mod.rs
```diff
@@ -155,6 +155,10 @@ impl IndexRecord {
 
 pub trait Fields {
     fn get_field_values(&self) -> Vec<(&'static str, u64)>;
+
+    fn get_record_size(&self) -> u64 {
```
tkaitchuck, line 159: I don't think this should be u64, it is at most a usize or i32 based on how it is being used above

tkaitchuck (approved):

tkaitchuck (commented):
Where does 111 come from? is there a Java PR that defines it?

AnkurAnand11:
> Where does 111 come from? is there a Java PR that defines it?

We have hardcoded the attribute Id here to 111 since randomly generating it was causing issue while reading.

src/index/writer.rs
```diff
@@ -95,14 +99,54 @@ impl<T: Fields + PartialOrd + PartialEq + Debug> IndexWriter<T> {
...
+        let uid = 111;
+        let request = Requests::UpdateSegmentAttribute(UpdateSegmentAttributeCommand {
+            request_id: get_request_id(),
+            segment_name: segment_name.clone(),
+            attribute_id: uid,
+            new_value: T::get_record_size() as i64,
+            expected_value: i64::MIN,
+            delegation_token: token,
+        });
+        raw_client
```
tkaitchuck, line 137: It is using `.expect` to check there was no error sending the command, but the reply from the server is not being looked at. It is possible the server returned an error, or something like "wrong host".

src/index/writer.rs
```diff
@@ -95,14 +99,54 @@ impl<T: Fields + PartialOrd + PartialEq + Debug> IndexWriter<T> {
...
+            .await
+            .expect("get endpoint for segment");
+
+        let raw_client = factory.create_raw_client_for_endpoint(endpoint);
+        let token = controller_client
+            .get_or_refresh_delegation_token_for(stream.clone())
+            .await
+            .expect("controller error when refreshing token");
+        let segment_name = segment_name.to_string();
+        let uid = 111;
```
tkaitchuck, line 128: This should be made into a constant

tkaitchuck (changes requested):

tkaitchuck:
> > Where does 111 come from? is there a Java PR that defines it?
> 
> We have hardcoded the attribute Id here to 111 since randomly generating it was causing issue while reading.

These should be defined in the `Attributes` class in java.

src/index/writer.rs
```diff
@@ -10,16 +10,22 @@
...
 use bincode2::Error as BincodeError;
+use pravega_wire_protocol::commands::UpdateSegmentAttributeCommand;
+use pravega_wire_protocol::wire_commands::Requests;
 use snafu::{ensure, Backtrace, ResultExt, Snafu};
 use std::fmt::Debug;
 use std::marker::PhantomData;
+use tracing::info;
 
 const MAX_FIELDS_SIZE: usize = 100;
+pub const INDEX_RECORD_SIZE_ATTRIBUTE_ID: u128 = 80000000000000000000000000000011;
```
tkaitchuck, line 28: Is this number supposed to be in hex? Because it is not. (It should start with `0x`) Also use some `_` characters to make it more readable. Also add a comment for where it comes from.

src/index/writer.rs
```diff
@@ -95,14 +101,53 @@ impl<T: Fields + PartialOrd + PartialEq + Debug> IndexWriter<T> {
...
+            segment_name: segment_name.clone(),
+            attribute_id: INDEX_RECORD_SIZE_ATTRIBUTE_ID,
+            new_value: T::get_record_size() as i64,
+            expected_value: i64::MIN,
+            delegation_token: token,
+        });
+        raw_client.send_request(&request).await.map_or_else(
+            |e| panic!("failed to get reply: {}", e),
+            |r| info!("UpdateSegmentAttribute : {:?}", r),
+        );
```
tkaitchuck, line 141: This change does not address my comment. The method send_request returns a Result where the value is the reply message from the server. You need to do a match on the reply from the server to make sure it is a AttributeUpdated.

src/index/reader.rs
```diff
@@ -110,11 +117,50 @@ impl IndexReader {
...
+            Replies::SegmentAttribute(cmd) => {
+                if cmd.value == i64::MIN {
+                    info!("Segment attribute for record_size is not set.Falling back to default RECORD_SIZE = {:?}", RECORD_SIZE);
+                    RECORD_SIZE as usize
+                } else {
+                    cmd.value as usize
+                }
+            }
+            _ => {
+                info!("get segment attribute for record_size failed due to {:?}", reply);
```
tkaitchuck, line 153: Humm, this actually brings up an interesting edge case:
Suppose the user creates the reader before the writer. Then we would get a segment does not exist error. In this case we have a couple of options: We could fail and simply not allow creating the reader first.
To make sure the fallback value is correct, we should add a check on reader:350 to make sure the value read was actually the correct size.

tkaitchuck (approved):

